PR #91, opened 2014-10-04
Add implicit optional format param to routes

### LINKED ISSUE #90: Router should recognize routes containing "."
e.g.

``` rust
let mut server = Nickel::new();
server.post("/foo/:bar", foo_handler);
```

will return 404 when requesting 
http://localhost:3000/foo/bar.baz

Ryman:
*BREAKING CHANGE*

This will break code that tries to handle routes such as: '/foo/:bar.:baz'

You can now handle this with: '/foo/:bar'

When this route matches '/foo/cats.gif' then request.param("format") will be ".gif"

Fixed #90.

cc: @pzol @cburgdorf @SimonPersson 

Currently we're using `([,a-zA-Z0-9%_-]*)` to capture variables in a route. Journey (the Rails router) [uses `([^./?]+)`](https://github.com/rails/journey/blob/v1.0.4/lib/journey/path/pattern.rb#L89). 
I suggest using a blacklist, or we'll continue to see more bugs in this area, e.g. '/foo/:bar' won't match '/foo/你好' which seems a shame.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -301,12 +301,12 @@ impl<'a> Router {
 
 impl HttpRouter for Router {
     fn add_route(&mut self, method: Method, path: &str, handler: RequestHandler) {
-        static FORMAT_SEQ: &'static str = "(\\.:format)?";
+        static FORMAT_VAR: &'static str = ":format";
 
-        let with_format = if path.ends_with(FORMAT_SEQ) {
+        let with_format = if path.contains(FORMAT_VAR) {
             path.to_string()
         } else {
-            format!("{}{}", path, FORMAT_SEQ)
+            format!("{}(\\.{})?", path, FORMAT_VAR)
         };
 
         let matcher = path_utils::create_regex(with_format[]);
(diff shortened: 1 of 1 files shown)


Ryman:
This was never actually merged? Will need a rebase.

cburgdorf:
Oh noooees, what did I do back then :'(

Ryman:
Just looked into the history on this one, it did merge and then encountered a bad rebase or something? It gets nuked out during a PR from ME :disappointed: https://github.com/nickel-org/nickel.rs/pull/94, specifically this commit https://github.com/Ryman/nickel.rs/commit/f12ab9ade802ee3c200db0812b2ffda3d15069df. 

I guess git was overzealous when a file got moved and didn't warn me about a merge conflict, still kind of shocking to not have noticed (and the tests got nuked so nothing failed out). So definitely not a forgotten merge anyway @cburgdorf!

So, I'm going to re-add the changes in a new PR since rebasing this doesn't do anything as it's already in master's history.

PR #124, opened 2014-11-24
feat(nickel): Added a println of IP:Port when the server starts listening for requests.

jin:
Hello,

Rust beginner here, and I wonder why there isn't a printout of the `ip:port` when the server starts. I find this output to be useful (as with many other server frameworks) because you don't have to refer to the source file to know what port it is.

If the absence of this printout is intentional, feel free to close this PR.

Thanks!

### DIFF
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -208,6 +208,13 @@ impl Nickel {
         }
 
         self.middleware_stack.add_middleware(not_found_handler);
+
+        match port {
+            80u16 =>  println!("Listening on http://{}", ip),
+            _ =>  println!("Listening on http://{}:{}", ip, port),
+        }
+        println!("Ctrl-C to shutdown server");
+
         Server::new(self.middleware_stack, ip, port).serve();
     }
 }


ixjf:
I don't understand why this was ever merged into the framework - a library, in my opinion, should not ever output information that way. The user of the library should be responsible for handling errors and producing output based on the behavior of the API.

cburgdorf:
Most web server frameworks print something like "listening on address x.x.x.x:xxxx" when after they are bound to listen. Do you actually know a web server framework that *doesn't*?

Ryman:
I can understand why it's not ideal for some applications, but it seems useful to get up and running quickly.

@ixjf  Would you like to open an issue or PR for making the behavior optional?

cburgdorf:
:+1:  for making it optional

ixjf:
@cburgdorf Yes, you're right. I guess making the behavior optional would be the best as Ryman said. I'll create a pull request as soon as I can.

PR #135, opened 2015-01-20
[WIP] Make it build with Rust 1.0.0-alpha

rgs1:
WIP: don't pull yet 

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -1,3 +1,5 @@
+#![allow(unstable)]
+
 extern crate serialize;
 extern crate nickel;
 extern crate http;
@@ -9,10 +11,10 @@ use nickel::{
 };
 use nickel::mimes::MediaType;
 use std::io::net::ip::Ipv4Addr;
-use std::collections::TreeMap;
+use std::collections::BTreeMap;
 use serialize::json::{Json, ToJson};
 
-#[deriving(Decodable, Encodable)]
+#[derive(Decodable, Encodable)]
 struct Person {
     firstname: String,
     lastname:  String,
@@ -20,7 +22,7 @@ struct Person {
 
 impl ToJson for Person {
     fn to_json(&self) -> Json {
-        let mut map = TreeMap::new();
+        let mut map = BTreeMap::new();
         map.insert("first_name".to_string(), self.firstname.to_json());
         map.insert("last_name".to_string(), self.lastname.to_json());
         Json::Object(map)
@@ -43,7 +45,9 @@ fn main() {
     }
 
     // middleware is optional and can be registered with `utilize`
-    server.utilize(logger);
+    // issue #20178
+    let logger_handler: fn(&Request, &mut Response) -> MiddlewareResult = logger;
+    server.utilize(logger_handler);
 
     let mut router = Nickel::router();
 
@@ -51,29 +55,41 @@ fn main() {
         format!("This is user: {}", request.param("userid"))
     }
 
+    // issue #20178
+    let uhandler: fn(&Request, &mut Response) -> String = user_handler;
+
     // go to http://localhost:6767/user/4711 to see this route in action
-    router.get("/user/:userid", user_handler);
+    router.get("/user/:userid", uhandler);
 
     fn bar_handler(_request: &Request, response: &mut Response) {
         response.send("This is the /bar handler");
     }
 
+    // issue #20178
+    let bhandler: fn(&Request, &mut Response) = bar_handler;
+
     // go to http://localhost:6767/bar to see this route in action
-    router.get("/bar", bar_handler);
+    router.get("/bar", bhandler);
 
     fn simple_wildcard(_request: &Request, response: &mut Response) {
         response.send("This matches /some/crazy/route but not /some/super/crazy/route");
     }
 
+    // issue #20178
+    let shandler: fn(&Request, &mut Response) = simple_wildcard;
+
     // go to http://localhost:6767/some/crazy/route to see this route in action
-    router.get("/some/*/route", simple_wildcard);
+    router.get("/some/*/route", shandler);
 
     fn double_wildcard(_request: &Request, _response: &mut Response) -> &'static str {
         "This matches /a/crazy/route and also /a/super/crazy/route"
     }
 
+    // issue #20178
+    let dhandler: fn(&Request, &mut Response) -> &'static str = double_wildcard;
+
     // go to http://localhost:6767/a/nice/route or http://localhost:6767/a/super/nice/route to see this route in action
-    router.get("/a/**/route", double_wildcard);
+    router.get("/a/**/route", dhandler);
 
     // try it with curl
     // curl 'http://localhost:6767/a/post/request' -H 'Content-Type: application/json;charset=UTF-8'  --data-binary $'{ "firstname": "John","lastname": "Connor" }'
@@ -82,8 +98,11 @@ fn main() {
         format!("Hello {} {}", person.firstname, person.lastname)
     }
 
+    // issue #20178
+    let phandler: fn(&mut Request, &mut Response) -> String = post_handler;
+
     // go to http://localhost:6767/a/post/request to see this route in action
-    router.post("/a/post/request", post_handler);
+    router.post("/a/post/request", phandler);
 
     fn json_response(_request: &Request, _response: &mut Response) -> Json {
         let person = Person {
@@ -93,15 +112,21 @@ fn main() {
         person.to_json()
     }
 
+    // issue #20178
+    let jresponse: fn(&Request, &mut Response) -> Json = json_response;
+
     // go to http://localhost:6767/api/person/1 to see this route in action
-    router.get("/api/person/1", json_response);
+    router.get("/api/person/1", jresponse);
 
     // try calling http://localhost:6767/query?foo=bar
     fn query_handler(request: &mut Request, _response: &mut Response) -> String {
-        format!("Your foo values in the query string are: {}", *request.query("foo", "This is only a default value!"))
+        format!("Your foo values in the query string are: {:?}", *request.query("foo", "This is only a default value!"))
     }
 
-    router.get("/query", query_handler);
+    // issue #20178
+    let qhandler: fn(&mut Request, &mut Response) -> String = query_handler;
+
+    router.get("/query", qhandler);
 
     // try calling http://localhost:6767/strict?state=valid
     // then try calling http://localhost:6767/strict?state=invalid
@@ -114,7 +139,10 @@ fn main() {
         }
     }
 
-    router.get("/strict", strict_handler);
+    // issue #20178
+    let sthandler: fn(&mut Request, &mut Response) -> MiddlewareResult = strict_handler;
+
+    router.get("/strict", sthandler);
 
     server.utilize(router);
 
@@ -134,7 +162,10 @@ fn main() {
         }
     }
 
-    server.handle_error(custom_404);
+    // issue #20178
+    let custom_handler: fn(&NickelError, &Request, &mut Response) -> MiddlewareResult = custom_404;
+
+    server.handle_error(custom_handler);
 
     server.listen(Ipv4Addr(127, 0, 0, 1), 6767);
 }
--- a/src/favicon_handler.rs
+++ b/src/favicon_handler.rs
@@ -50,7 +50,7 @@ impl FaviconHandler {
     #[inline]
     pub fn is_favicon_request (req: &request::Request) -> bool {
         match req.origin.request_uri {
-            AbsolutePath(ref path) => path.as_slice() == "/favicon.ico",
+            AbsolutePath(ref path) => &path[] == "/favicon.ico",
             _                      => false
         }
     }
@@ -74,8 +74,8 @@ impl FaviconHandler {
     }
 
     pub fn send_favicon (&self, req: &request::Request, res: &mut response::Response) {
-        debug!("{} {}", req.origin.method, self.icon_path.display());
+        debug!("{:?} {:?}", req.origin.method, self.icon_path.display());
         res.content_type(MediaType::Ico);
-        res.send(self.icon.as_slice());
+        res.send(&self.icon[]);
     }
 }
--- a/src/json_body_parser.rs
+++ b/src/json_body_parser.rs
@@ -1,17 +1,16 @@
 use std::str;
 use serialize::Decodable;
-use serialize::json::{Decoder, DecoderError};
 use request::Request;
-use typemap::Assoc;
-use plugin::{Phantom, PluginFor, GetCached};
+use typemap::Key;
+use plugin::{Phantom, Plugin, Pluggable};
 
 // Plugin boilerplate
 struct JsonBodyParser;
-impl Assoc<String> for JsonBodyParser {}
-impl<'a, 'b> PluginFor<Request<'a, 'b>, String> for JsonBodyParser {
+impl Key for JsonBodyParser { type Value = String; }
+impl<'a, 'b> Plugin<Request<'a, 'b>> for JsonBodyParser {
     fn eval(req: &mut Request, _: Phantom<JsonBodyParser>) -> Option<String> {
         if !req.origin.body.is_empty() {
-            str::from_utf8(req.origin.body.as_slice()).map(|s| s.to_string())
+            str::from_utf8(&*req.origin.body).ok().map(|s| s.to_string())
         } else {
             None
         }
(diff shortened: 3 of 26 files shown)


rgs1:
Not ready yet, but more progress

SimonPersson:
I saw that the dependencies broke today again. I updated them so we can keep working on this. Are you still working on this @rgs1? If not I might try to finish it during the weekend.

dkhenry:
Ok so @rgs1 I have fixed most the errors from the latest nightly, but I have no idea what the following construct's are doing so I don't even know how to begin to fix them. 

    json_body_parser.rs:34:16: 34:30 error: type parameters may not appear here [E0085]
    json_body_parser.rs:34         self::<JsonBodyParser>().and_then(|&: parsed| {

If you can point me in the direction of what that language construct even is I can have a pull request for you that should fix all the build errors.

SimonPersson:
You must have edited that line.

What it originally says is `self.get::<JsonBodyParser>().and_then(|parsed|`.

What that is supposed to do is the following:
* Retrieve the `JsonBodyParser` from the typemap.
* If it was found, execute the function on the following line.
   * If it wasn't found, return without doing anything.

I assume the confusion comes from the fact that you have edited the line.

I'm super glad to see new contributors by the way!

rgs1:
Ah very nice @dkhenry! I'll resume this today!

BenMorganIO:
:+1:

rgs1:
Ok! It builds! It's not ready for merge because there are some inefficient things (i.e.: assigning a Regex::new to a static inside a mod is not working for me, need to dig in) plus in general things are not super tidy. I also haven't touched the tests. 

I'll make a 2nd pass now to get things in order, unless someone gets there first.

rgs1:
(I update the branch, which is till WIP)

rgs1:
Ah, so regex! is available via a separate lib (plugin), regex_macros. So the static issue is fixed, next warnings + tests!

rgs1:
No more warnings! Only tests are left to be fixed!

SimonPersson:
Beautiful! I will have a look at the code later (as will the others, I'm sure. @Ryman always manage to find some nitpicks :) ). When you feel that you're done, would you mind naming your commit according to [the following conventions?](https://github.com/ajoslin/conventional-changelog/blob/master/CONVENTIONS.md) I forgot to tell you on your last commit but we're looking at automatically generating a changelog in the future.

rgs1:
Tests build!! They don't pass though :-) But I think I mostly know why. Going out for a run and will fix them when I get back. Almost there! Stay tuned

SimonPersson:
You have changed the mutability on the `Request` parameter to some handler functions in the examples where it isn't necessary. Interestingly enough, the compiler doesn't warn us that the mutability isn't needed. Anyways, as we don't need it to be mutable in these examples we might just as well change that back.

And what's up with storing the handler in a variable like that? Are you sure issue #15444 is related? Because this was working a month and a half ago and that issue is older, no?

rgs1:
Oops, sorry about the unneeded mut. Sure I'll fix all the unneeded stuff on a final pass (making tests pass now).

Wrt to coercing handlers, I couldn't find another way to make it work with nightly. Here's a toy example: http://is.gd/H60Xjl.

I asked around #rust, and they say it'll get fix at some point.

SimonPersson:
No worries. Sounds good!

Yeah, I also figured that it must be a bit of a bug in the compiler. I questioned whether issue #15444 that you mentioned in a comment is the one to blame, as that issue dates from before this stopped working. Anyways, we have the `router!` macro available so people won't have to use that workaround if the don't want to :)

We'll have to remember to revert the examples once this gets fixed in the compiler just.

rgs1:
Yup I'll try to find the actual real/current issue in so we can mention that in the FIXMEs

rgs1:
All tests passing! Now, we can do some real review @SimonPersson  ;-)

Up next:

* drop the unneeded &mut 
* format the commit according to the contributing guidelines

Please feel free to point out nits (and not nits) so we can get this in good shape. Thanks!

rgs1:
Updated the commit message according to the used convention + drop the unneeded &mut changes. Should be ready-ish for merge hopefully.

SimonPersson:
Yay! Travis passed too, great news.

The `not_found_handler` in the file nickel.rs is commented out (line 142). Is there a reason for that or did you just miss it?

Some calls to `unwrap()` are introduced. These are not good, as they can cause the server to panic. For example, line 108 in `response.rs` was changed from using `and_then` to `unwrap`. Can't we use `and_then` for some reason here? If we can't use `and_then` for some reason, I suppose we can use normal pattern matching. Could you look at all the calls to `unwrap()` that you introduced and see if they can be eliminated (preferably with an error message on the failing case where it makes sense, like in the creation of regexes from handler path strings)? I think it's preferable that the server is misbehaving on incorrect data rather than crashing.

Line 140 in `response.rs` was moved, now the comments there doesn't make much sense. Line 139 and 140 should be switched.

Other than these things, I think this looks great! I'd like to hear the opinion of @Ryman or @cburgdorf before merging though,

rgs1:
Great, thanks for the review! I'll take a look in a bit, all of those are probably left over or omissions.

rgs1:
Fixed the accidentally commented not_found_handler in src/nickel.rs. Will get to the unwrap() calls next.

rgs1:
Fixed all unwrap calls to do full matching instead (except inside tests, where I guess it's ok). Not sure I do the right thing on some Err cases, feedback welcomed!

SimonPersson:
This looks good to me. I don't think there is a good way to handle errors all the time, we just need to do something so that the server won't crash on malformed input.

Some more small things:

* In the commit comment, you have written "Use BTreeMap instead of BTreeMap", that must be a mistake.
* There's still one unnecessary `mut` in  macro_example.rs.
* The `not_found_handler` in nickel.rs should have only `_` as parameter names as they aren't used.

Other than that, I think this is ready for merging. Unless the others find something bad, that is.

dkhenry:
I don't know if its a problem with the recent changes or If my client program is bad, but trying to test it out with wget is giving me 
    2015-01-26 07:57:28 ERROR -1: Malformed status line.

The driver is pretty much the example driver from the documentation.

SimonPersson:
That's interesting! I don't get it with my browser, neither with curl but wget gives that error. We have to investigate this. It could be that we've messed up rust-http somewhere.

SimonPersson:
The problem seems to be that we don't send the status code with the response. The response looks like this:
```
HTTP/1.1 Ok
```
When it should have been:
```
HTTP/1.1 200 Ok
```

Wget complains about this while curl and the browser just pretends like nothing is wrong.

SimonPersson:
I've just confirmed that this happens in the rust-http examples too. Lets continue the investigation in that repo.

rgs1:
Updated the commit according to @SimonPersson's feedback.

rgs1:
Wrt to the rust-http issue, in https://github.com/nickel-org/rust-http/commit/53d30c0d74875f3db502ba8c330bfdbf42b0db16 I wonder if:

```rust
-impl fmt::Show for Status {
+impl fmt::Display for Status {
fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
    write!(f, \"{} {}\", self.code(), self.reason())
}
```

has something to do.

examples/example.rs
```diff
@@ -43,7 +45,9 @@ fn main() {
     }
 
     // middleware is optional and can be registered with `utilize`
-    server.utilize(logger);
+    // issue #15444
+    let logger_handler: fn(&Request, &mut Response) -> MiddlewareResult = logger;
+    server.utilize(logger_handler);
```
Ryman: `server.utilize(logger as fn(&Request, &mut Response) -> MiddlewareResult);` is the common form of this last I saw, but it's still horrible :(

examples/example.rs
```diff
@@ -51,29 +55,41 @@ fn main() {
         format!("This is user: {}", request.param("userid"))
     }
 
+    // issue #15444
```
Ryman: I don't think this relates to #15444, it's more similar to #20178?

src/favicon_handler.rs
```diff
@@ -50,7 +50,7 @@ impl FaviconHandler {
     #[inline]
     pub fn is_favicon_request (req: &request::Request) -> bool {
         match req.origin.request_uri {
-            AbsolutePath(ref path) => path.as_slice() == "/favicon.ico",
+            AbsolutePath(ref path) => &path[0..] == "/favicon.ico",
```
Ryman: I think all instances of `foo[0..]` should just be `foo[]` ?

src/json_body_parser.rs
```diff
@@ -1,17 +1,19 @@
...
 // Plugin boilerplate
 struct JsonBodyParser;
-impl Assoc<String> for JsonBodyParser {}
-impl<'a, 'b> PluginFor<Request<'a, 'b>, String> for JsonBodyParser {
+impl Key for JsonBodyParser { type Value = String; }
+impl<'a, 'b> Plugin<Request<'a, 'b>> for JsonBodyParser {
     fn eval(req: &mut Request, _: Phantom<JsonBodyParser>) -> Option<String> {
         if !req.origin.body.is_empty() {
-            str::from_utf8(req.origin.body.as_slice()).map(|s| s.to_string())
+            match str::from_utf8(&req.origin.body[0..]) {
```
Ryman: `str::from_utf8(&*req.origin.body).ok().map(|s| s.to_string())`

src/favicon_handler.rs
Ryman: Alternatively `&*foo`, but perhaps slicing syntax is more explicit (I haven't checked the recent style guide on this)

src/middleware_handler.rs
```diff
@@ -84,7 +85,7 @@ impl ResponseFinalizer for json::Json {
     }
 }
 
-impl<'a, S: Show> ResponseFinalizer for &'a [S] {
+impl<'a, S: Display> ResponseFinalizer for &'a [S] {
```
Ryman: nit: the commit message calls this 'Derive' instead of 'Display'

src/nickel_error.rs
```diff
@@ -27,10 +28,10 @@ impl NickelError {
     }
 }
 
-#[deriving(Show)]
+#[derive(Show)]
 pub enum NickelErrorKind {
     // FIXME: Should probably re-export http::status::Status
     ErrorWithStatusCode(Status),
-    UserDefinedError(int, String),
+    UserDefinedError(usize, String),
```
Ryman: This is a semantic change, but I'm mostly alright with it. The error system still needs a whole revamp to match modern conventions.

src/response.rs
```diff
@@ -106,8 +106,7 @@ impl<'a, 'b> Response<'a, 'b> {
         self.origin.headers.content_length = None;
 
         self.origin.headers.content_type = path.extension_str()
-                                               .and_then(from_str)
-                                               .map(mimes::get_media_type);
+            .and_then(|s| s.parse()).map(mimes::get_media_type);
```
Ryman: nit: personally consider old formatting more skimmable

src/response.rs
```diff
@@ -124,46 +123,52 @@ impl<'a, 'b> Response<'a, 'b> {
...
+    pub fn render<'c, T: Encodable>
         (&mut self, path: &'static str, data: &T) {
             // Fast path doesn't need writer lock
-            match self.templates.read().get(&path) {
-                Some(t) => {
-                    let _ = t.render(self.origin, data);
-                    return
+            match self.templates.read() {
+                Ok(hm) => match hm.get(&path) {
+                    Some(t) => {
```
Ryman: `if let Some(t) = gm.get(&path) { .. }`

src/response.rs
```diff
@@ -124,46 +123,52 @@ impl<'a, 'b> Response<'a, 'b> {
...
+                        let _ = t.render(self.origin, data);
+                        return
+                    },
+                    None => {}
                 },
-                None => {}
+                Err(_) => {}
```
Ryman: Probably shouldn't eat this error without writing it somewhere

src/response.rs
```diff
@@ -124,46 +123,52 @@ impl<'a, 'b> Response<'a, 'b> {
...
+                        },
+                        Occupied(entry) => entry.into_mut()
+                    };
+                    let _ = template.render(self.origin, data);
                 },
-                Occupied(entry) => entry.into_mut()
-            };
-
-            let _ = template.render(self.origin, data);
+                Err(_) => {}
```
Ryman: Same here

src/router/mod.rs
```diff
@@ -34,20 +34,24 @@ mod path_utils {
...
 
-    pub fn get_variable_info (route_path: &str) -> HashMap<String, uint> {
+    pub fn get_variable_info (route_path: &str) -> HashMap<String, usize> {
         REGEX_VAR_SEQ.captures_iter(route_path)
              .enumerate()
-             .map(|(i, matched)| (matched.at(1).to_string(), i))
+            .map(|(i, matched)|
+                 match matched.at(1) {
+                     Some(m) => (m.to_string(), i),
+                     None => ("".to_string(), i)  // better way to handle the error?
```
Ryman: Change the `map` to `filter_map` to drop the `None`s perhaps.

Ryman:
Added a few inline comments, looks good other than that, and we can iterate easier once master's buildable :) I've let an old branch porting to hyper bitrot, might try to rebase onto this sometime soon. 

Thanks for the efforts @rgs1!

SimonPersson:
@Ryman - Hyper sounds cool (and needed)!

@rgs1 - I pushed a fix to the http repo, reverting your changes in that file. Now it works again :)

src/response.rs
SimonPersson: It says in the documentation that this only occurs if the writer panics while an exclusive lock, in that case we're screwed and might as well panic here too. Unwrap might not be so bad.

@rgs1 I'm sorry I made you change the `unwrap()` calls yesterday, I didn't read carefully enough last night.

rgs1:
@SimonPersson: thanks! Though Show is deprecated and was split into Debug and Display. I think we want both for status (i.e.: impl Display and derive Debug). We can follow-up in that repo.

@Ryman: thanks for the review! I'll get to that a bit later, would be great to have master be buildable again!

examples/example.rs
rgs1: So I guess we leave it as is and revisit later on when the variants of https://github.com/rust-lang/rust/issues/20178 are done?

examples/example.rs
rgs1: Updated

src/favicon_handler.rs
rgs1: Updated to []

src/json_body_parser.rs
rgs1: Changed

src/middleware_handler.rs
rgs1: Oops, fixed

src/response.rs
rgs1: ah, sorry for breaking the style - novice mistake. fixed.

src/response.rs
rgs1: Changed

src/response.rs
rgs1: So calling unwrap() on read() and write() then

src/response.rs
rgs1: Changed

src/router/mod.rs
rgs1: Fixed

rgs1:
Updated to conform with @Ryman's review - let me know how it looks now.

examples/example.rs
Ryman: I think it's worth trying:
`type HandlerPtr = fn(&Request, &mut Response) -> MiddlewareResult`
and changing fallout code to:
`foo(bar as HandlerPtr)`

*OR* I think changing the naming convention, `logger_handler` => `logger_fnptr`.

I guess it's a nit really, but I'm guessing people may understand a cast to a function pointer if they come from c++/c rather than reading 'I'm turning my handler into a handler..?' and perhaps being confused.

cc @SimonPersson

Ryman:
@rgs1 Thanks, I think I checked all the changes, still a few outstanding nits. 

In future, could you make changes in new separate commits that can be squashed later, as it helps to minimize incremental review time. :+1:

rgs1:
Added two commits to address the last two comments. Shall we wait for @SimonPersson to weight in on how to handle the casts?

Note that the last two commits don't adhere to the usual commit formatting, since we'll probably squash them. 

Thanks for the review @Ryman.

Ryman:
LGTM once there's a consensus on the casting/ergonomics issue :)

SimonPersson:
Have we tried declaring a type for easier casting? If it works it is prettier I suppose. On the other hand, it's not meant to be a permanent solution anyways and there's still the `router!`-macro for those who value ergonomics.

So, it doesn't really matter to me :)

rgs1:
I assume we'll be tracking nightly from now on, so things will be in a state of flux anyway. As soon as this is fixed in the compiler, we can change it.

SimonPersson:
Exactly. I say we merge this as it is.

@rgs1 - Can you squash the commits to make it ready for merging?
@Ryman - You press the button if you can't think of anything more!

rgs1:
Squashed @SimonPersson @Ryman !

thiagopnts:
Awesome work @rgs1 :)

Ryman:
Merged, thanks for the work @rgs1! :+1:

PR #143, opened 2015-02-07
fix(*): std::path is now std::old_path

rgs1:


SimonPersson:
I already made a pull request with this change, #141. As no one merged that one yet, I'll go ahead and merge this so we get a fix.

PR #151, opened 2015-02-16
chore(time, url): use dependencies from crates.io

rgs1:
Since yesterday, rust-http is using deps from crates.io instead of git: https://github.com/nickel-org/rust-http/commit/22cb5e85ecea0cfe4f09440f8d58671c30642bde.

So builds are broken because the definitions of struct time::Tm in git and crates.io don't match. The solution is for nickel to use the crates.io deps as well. 


Ryman:
Already pushed a similar change to master last night, closing. Thanks for taking the time though!

PR #157, opened 2015-02-27
feature(hyper): <insert *fireworks* here>

### LINKED ISSUE #99: Response#send should handle state
Currently multiple calls to send are allowed, but each call will try to set headers again. We should probably give Response a state, so we know if the headers have been written or not. We can additionally use this information in the finalizers, if nothing has been written to the response then we can add some additional headers like `Content-Length` if it's simple to calculate the length.

How we handle this would also affect things like `res.content_type(..)` since we shouldn't really allow the user to modify headers after already trying to send data. For example, it is a bug for a user to `send` and then call `status_code`, we should help them catch this and not have to debug through curl.

We could achieve this through the type system via Phantom types, meaning no runtime cost, but it will make current finalizers pretty broken. *But*, I think we can probably just have it so that `RequestHandler`s don't have a response as their input for simple cases which would make it less of a problem.

e.g. This would be a `200 OK` with a content length automatically set
````
fn handler(req: &Request) -> String {
    format!("{}", req.param("something"))
}
````
I think this could actually make things somewhat easier to test for users too. (change &Request to some Request trait)

The full handler form should still be available for whenever a response needs to be progressively generated and chunked to the client.

If we chose to manage state at runtime we could still do the above, the only question is what to do if a user calls something they shouldn't do after a `send`, do we fail? print a warning? just no-op?

An alternative could be to add a lint that helps find method calls (other than send) on a response after a send has been performed.

Ryman:
This is basically feature parity with master, there's still a decent amount I'd like to change/introduce, but I feel this is in a state where it should be reasonably stable enough for review and can allow us to iterate together.

This obviously breaks a lot of code, and I'm especially unsure about the error handling. I've chosen to only allow error handlers to operate on a response that is already in progress (I still think `NickelError` needs a redesign but didn't want to sneak an opinionated rewrite into this). The reasoning is that either the handler knows what the error is (`BadRequest`, etc) and can give appropriate status and kick off the request or the error happens mid-stream and it would be a logic error to have an error handler try to edit the `StatusCode` mid-response. The `stream` is optional in `NickelError` as there could be a write error, and then writing on the stream is useless, and the handlers point is just to log.

Most blanket impls for `Middleware` are removed so that we only have one blanket impl (which will be a requirement for unboxed closures). I actually left in two since one is equivalent to an unboxed closure with ref captures which can be removed in future. 

I added convenience macros `middleware!` to provide similar syntax as an unboxed closure as there are still tricky lifetime issues that need to be dealt with to get them working (I gave it a number of frustrating attempts before moving on with a *more flexible* macro based approach). I think it might actually be worth adding an `error_handler!` variant and splitting the current macro into `middleware!` and `handler!`, the former returning `MiddlewareResult`, the latter `T: ResponseFinalizer`.

Documentation has only really been updated to pass tests, a lot of it could benefit from a rewrite, I don't feel it's a blocking issue.

Bare in mind this has been through a number of rebases, so please point out nits where I've botched something! Also open to naming bikeshedding for anything introduced here. I'll probably do some additional cleanup in the next day or two.

cc @SimonPersson @cburgdorf @rgs1 

This will fail travis until typemap & plugin merge upstream PRs.

Closes #99.


SimonPersson:
Cool! It wasn't as big of a change as I anticipated really, no huge architectural changes, so that's nice.

I left a few comments but it looks nice overall. Super exciting news this! Thanks for the great work @Ryman!

rgs1:
generally lgtm! having this merged sooner rather than later (even with some rough edges) would be nice so that catching up with Hyper is simpler and so that we can start using it. thanks @Ryman !

Ryman:
Lots of upstream change to new `io` on todays nightly so rust-http is now completely broken :( 

I tried to fix it up but each fix just leads to more compiler errors as it unveils more typechecking issues etc. Hyper's broken too but they have a patch already done that just needs to land, which then probably needs adjustment on this branch to accommodate but then we're good?

rgs1:
yeah if we merge this, we can then all pitch in for polishing the rough edges and keeping up to date with hyper.

ninjabear:
@rgs1 - seconded

also @Ryman, nice one!

steveklabnik:
This would be neat becuase currently, nickel fails to build due to rust-http

Ryman:
@SimonPersson Pushed the `io` fixes and I think your comments are addressed? Looks like the latest nightly (even though I downloaded mine like 2hours ago..) is broken for some upstream stuff though so Travis is still going to fail ( https://github.com/rust-lang/regex/pull/51 )

SimonPersson:
Looks nice to me!

rgs1:
w00t! are we ready to rock/merge? :-)

Ryman:
@SimonPersson upstream's fixed, travis passing, merging!

SimonPersson:
Yay! Great work @Ryman!

rgs1:
awesome - thanks @Ryman !

PR #158, opened 2015-03-05
fix(middleware_handler): use from_usize instead of from_uint

rgs1:
Remove warning.


Ryman:
Thanks!

Issue #162, opened 2015-03-05
Static file MIME type inferred incorrectly for 404s


ninjabear:
As I alluded to before without explaining properly, what I mean is it looks like the mime type is inferred from the file extension, but this should be overridden on certain events. 

If you load up nickel-bootstrap and put in /images/afilethatdoesntexist.jpg instead of getting a 404 I get a firefox (YMMV) "image cannot be displayed because it contains errors" response. 

Above is all speculation so far - it could be something else entirely. It could also be a hyper issue, depending on where the handover is. 

Interestingly this doesn't occur in chrome, which displays the correct "file not found" response.

Ryman:
Okay, it's 2 issues overlapping:

First, #163, mybad.

Secondly, The route "**" should match anything, but it's not doing so in this case and falling back to the default 404 handler instead.  This is due to the file extension, which I fixed a long time ago, and when checking the code... it had vanished... so it looks like #91 was closed rather than merged accidentally and it wasn't noticed >_<

I'll look at rebasing that soon if noone else beats me to it!

Ryman:
@ninjabear Can you confirm this is fixed?

ninjabear:
@Ryman hum, I can't really test at the moment, not sure if its the serve static directories fix or something else but nickel-bootstrap just hangs on everything, will have to look properly later

SimonPersson:
I'm seeing this issue. If I don't set the content type manually when responding with `response.send()` the response doesn't get a content type:
```
HTTP/1.1 200 OK
Date: Tue, 10 Mar 2015 12:51:28 GMT
Server: Nickel
Transfer-Encoding: chunked

Lalalalala
```
Firefox then assumes the response is an image if the requested url ends with `.jpg` and you get the "image cannot be displayed because it contains errors" message in the browser. My workaround for now is to set the response type manually when sending text on requests that could be an image.

ninjabear:
I looked into it a bit further and my problem could be related to multiple gets with a keep-alive, as all the URLs work individually but not the landing page at once.

Ryman:
@SimonPersson Got it, we should set default mime to PlainTxt or Html even for those handlers (I think I removed it at one point when implementing the ResponseFinalizers). 

@ninjabear The keepalive issue may be related to https://github.com/hyperium/hyper/issues/368 ?

SimonPersson:
Yeah, a default mime should do it.

[...]

PR #163, opened 2015-03-05
fix(mime): set Content-Type header when appropriate

Ryman:
Embarrassing cause of one of the problems in #162


Ryman:
Added the additional fix required for #162, r? @SimonPersson

SimonPersson:
Great. Merged.

PR #166, opened 2015-03-08
fix(StaticFilesHandler): don't try to send dirs

rgs1:
A follow-up feature would be to support directory
listings from StaticFilesHandler.


Ryman:
Thanks

PR #168, opened 2015-03-11
Default Content-Type & Temp fix for send_file deadlocking on non-existant files

Ryman:
cc #162, @SimonPersson, @ninjabear 

The 2nd commit is really just a temporary fix, we will need to adjust the `Response` api to deal with a different error type rather than `io::Result` otherwise things like this won't propagate to error handlers and it will be easy to drop the Response without flushing the buffer (which is causing the deadlocks). Hopefully deadlocks die off (no pun intended) once the 'new io' has support for timeouts, but for now we're going to have to develop an api that can hopefully ensure that it is impossible to lose the `Response` in a handler without a panic.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,6 @@
 #![crate_name = "nickel"]
 #![crate_type = "rlib"]
-#![feature(plugin, core, collections, io, net, path, path_ext, std_misc, old_path)]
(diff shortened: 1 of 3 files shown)


SimonPersson:
I don't have to much time right now, but it looks good from what I can see :)

PR #170, opened 2015-03-16
Refactor Response API and trim NickelError

Ryman:
Alternative to #168. 

Providing a separate PR incase the design isn't agreeable (this is based on that branch). 

These changes make it easier to enforce standard behavior. HTTP doesn't really allow changing the status code once a response is already in progress, so users should either bail (and log) or start an error with an appropriate code which error handlers can process. 

I considered changing the API for an ErrorHandler to take `&mut Option<Response<Streaming>>` but decided to keep `NickelError` as in the absense of a `Response` we can at least still know what the developers error message was (in case they want to log *something* to a database rather than *nothing*).

PR #171, opened 2015-03-18
Regex routes

### LINKED ISSUE #101: Regex Routing
What was the motivation for choosing a wildstar / custom routing language over matching paths with regex?  Is there any interest in adding support of regex path matching?  I know that personally I prefer using Regex to do my routing.  It provides more control than wildstar/blobs, and named capture groups match up with url parameters really well.

Or maybe Nickel.rs can already do this and I'm just not reading close enough.

If people are interested in using regex for routing, I'd be happy to patch it in.

Ryman:
This is based on top of #170 so will need rebasing for any changes there, alternatively I can rebase on master if that gets rejected. 

The last 3 commits are the ones worth looking at.

Closes #101.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,12 +18,12 @@ time = "*"
 typemap = "*"
 plugin = "*"
 regex = "*"
-regex_macros = "*"
 rustc-serialize = "*"
 log = "*"
 hyper = "*"
 groupable = "*"
 mustache = "*"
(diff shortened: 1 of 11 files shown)


Ryman:
Rebased and addressed previous feedback

Ryman:
Merging this as it allows us to build on beta and I'd rather not leave it stale any longer.

PR #173, opened 2015-03-26
fix(rustup): forced feature gates and generic conversions landed

Ryman:
Waiting on unsafe-any and typemap to get new versions published, so this will fail travis until later.


Ryman:
Pending on hyper

PR #176, opened 2015-04-01
Update to newest Hyper.

mattyhall:
Hyper now requires the Date header to be a Tm wrapped in a HttpDate.


mattyhall:
Oops. Didn't see #175. I'm an idiot!

Issue #120, opened 2014-11-21
Publish on crates.io


cburgdorf:
In order to get nickel listed on cargo.io, we need to release a first version. Any hesitations to release a 0.1.0 from current master? @Ryman @SimonPersson ?

hjr3:
This is blocking me from using postgres with nickel. The same openssl library is being sourced from crates.io and directly from a git repository.

Cargo.toml
```toml
[package]

name = "example"
version = "0.1.0"

[[bin]]

name = "example"
path = "src/main.rs"

[dependencies.nickel]

git = "https://github.com/nickel-org/nickel.rs.git"

[dependencies.postgres]

git = "https://github.com/sfackler/rust-postgres.git"
```

Output:

```
$ cargo update
    Updating git repository `https://github.com/nickel-org/nickel.rs.git`
    Updating git repository `https://github.com/sfackler/rust-postgres.git`
    Updating git repository `https://github.com/reem/rust-plugin`
    Updating git repository `https://github.com/nickel-org/rust-url.git`
    Updating git repository `https://github.com/rust-lang/time`
    Updating git repository `https://github.com/nickel-org/rust-http.git`
    Updating git repository `https://github.com/nickel-org/groupable-rs`
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating git repository `https://github.com/nickel-org/rust-mustache.git`
    Updating git repository `https://github.com/nickel-org/anymap.git`
    Updating git repository `https://github.com/sfackler/rust-openssl.git`
$ cargo build
 Downloading phf_mac v0.4.4
 Downloading time v0.1.7
 Downloading regex v0.1.6
 Downloading typemap v0.0.6
 Downloading phf v0.4.4
 Downloading gcc v0.1.3
 Downloading phf_shared v0.4.4
 Downloading log v0.1.7
 Downloading openssl v0.2.11
 Downloading openssl-sys v0.2.11
 Downloading uuid v0.1.4
 Downloading pkg-config v0.1.3
 Downloading rustc-serialize v0.2.2
 Downloading unsafe-any v0.2.0
native library `openssl` is being linked to by more than one package, and can only be linked to by one package

  openssl-sys v0.2.11
  openssl-sys v0.2.11 (https://github.com/sfackler/rust-openssl.git#df9a7357)
```

rgs1:
I guess the big blocker here would be moving out of rust-http (no longer maintained) and onto hyper?

ninjabear:
[...]

PR #177, opened 2015-04-03
fix(rustup): Coherence and Error changes

Ryman:
Also now using groupable from `crates.io`


Ryman:
Should be able to publish after `mustache` is pulled and published, cc #120

PR #187, opened 2015-04-13
fix(macros): remove nickel_macros crate

### LINKED ISSUE #183: can't find crate for `nickel_macros`
after installing nickel.rs as a Cargo crate, I can't compile code from examples. Always error `can't find crate for 'nickel_macros'`

nickel 0.2.0
rust 1.0.0 beta

Ryman:
The original separation was due to the instability of macros and the
potential use of a compiler plugin, but now that the beta has landed
we can have macros in the main crate knowing they are stable. Users
can opt-in to using the macros via the `macro_use` attribute.

[breaking-change]

Closes #183


Ryman:
r? @cburgdorf 

This should fix #183

cburgdorf:
Ah, perfect! Let's do that. The change looks good to me.

cburgdorf:
Btw, any chance to get you over to Gitter: https://gitter.im/nickel-org/nickel.rs
I'd like to chat with you if you have some minutes :)

Issue #189, opened 2015-04-14
Enchance regex route tests


lambdaburrito:
I've been debugging a routing issue and I though it would be good to start adding more regex route tests as the currents ones only match numbers (`4711`, `5490`) and not character-based variable paths. It would be ideal to test if hyphens work too.

I'm submitting a pull request for this.

Ryman:
Closed via #215

PR #190, opened 2015-04-14
enhance regex route matches

lambdaburrito:
this contributes to #189

### DIFF
--- a/src/router/router.rs
+++ b/src/router/router.rs
@@ -159,23 +159,31 @@ fn creates_valid_regex_for_routes () {
     assert_eq!(regex1.is_match("foo/4711/bar?foo=true&bar=false"), false);
     assert_eq!(regex1.is_match("foo/4711/bar/test%20spacing"), true);
     assert_eq!(regex1.is_match("foo/4711/bar/5281?foo=test%20spacing&bar=false"), true);
+    assert_eq!(regex1.is_match("foo/alice/bar/bob"), true);
 
     assert_eq!(regex2.is_match("foo/4711/bar"), true);
     assert_eq!(regex2.is_match("foo/4711/barr"), false);
     assert_eq!(regex2.is_match("foo/4711/bar?foo=true&bar=false"), true);
     assert_eq!(regex2.is_match("foo/4711/4712/bar"), false);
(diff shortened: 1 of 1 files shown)


src/router/router.rs
```diff
@@ -159,23 +159,31 @@ fn creates_valid_regex_for_routes () {
...
     assert_eq!(regex1.is_match("foo/4711/bar/5281?foo=test%20spacing&bar=false"), true);
+    assert_eq!(regex1.is_match("foo/alice/bar/bob"), true);
 
     assert_eq!(regex2.is_match("foo/4711/bar"), true);
     assert_eq!(regex2.is_match("foo/4711/barr"), false);
     assert_eq!(regex2.is_match("foo/4711/bar?foo=true&bar=false"), true);
     assert_eq!(regex2.is_match("foo/4711/4712/bar"), false);
     assert_eq!(regex2.is_match("foo/4711/4712/bar?foo=true&bar=false"), false);
+    assert_eq!(regex2.is_match("foo/alice/bar/bob"), false);
+    //assert_eq!(regex1.is_match("foo/alice/bar"), true);
```
Ryman: Should this be uncommented? Also perhaps you could extend the route tests to assert the data captured is what is expected? 

Thanks for noticing the gaps in the tests!

lambdaburrito: It's commented because it fails as it doesn't match and not sure if it
should or not?

Ryman: Should it be `regex2.is_match(..)` rather than `regex1`?

lambdaburrito: Good spot. What a noob mistake ^-^

Good idea to assert the data captured too. Want me to do this too?

Ryman: Would be neat if you could!

Also, mind rebasing to drop the merge commit and to adjust the message into line with [convention](https://github.com/ajoslin/conventional-changelog/blob/master/CONVENTIONS.md)?

lambdaburrito: No problem. ^-^ I'm new to GitHub so I'm not familiar with conventions and
best practises.

I will aim to get this in by the end of Friday.

Ryman: Any update on this? :)

Ryman:
Closed via #215

PR #192, opened 2015-04-16
Fix custom error handlers not being called

### LINKED ISSUE #191: Custom error handler is never called
I have debugged to discover that a custom error handler is never called!
I was wondering why my custom 404 handler wasn't return the correct response and digged further to find it's not even be called.

Ryman:
Closes #191


Ryman:
@lambdaburrito can you confirm if this fixes your issue?

PR #195, opened 2015-04-16
fix(query): allow multiple query params alive at once

Ryman:
I'd like to bikeshed on this API a bit.

Due to requiring a mutable reference to update the cached value, we can't return a view from `query`. This changes it to return a `store` which then you can then query into as much as you want without borrowck problems. An alternative is to keep the current api and then have a second method which returns the store like this currently does and maintaining the old api for the simple/average case.

I'm actually more in favor of changing the API a bit further, as I think most users will be wanting the first value in the set of matches so we should make that nicer.

```rust
let q : Query = req.query();
let bar : Option<&str> = q["bar"];
let bar : &str = bar.unwrap_or("default");

// When it only needs to live for the call
let bar : Option<&str> = req.query()["bar"];

// Alternatively this could return an empty slice instead of `None`?
let all_bar : Option<&[String]> = q.all("bar");
let all_bar : Vec<String> = all_bar.map(|v| v.to_vec()).unwrap_or(vec![]);
```

I think this would be more practical for most uses while not being too horrible for the extended case.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -85,13 +85,13 @@ fn main() {
     // try calling http://localhost:6767/query?foo=bar
     router.get("/query", middleware! { |request|
         format!("Your foo values in the query string are: {:?}",
-                request.query("foo", "This is only a default value!"))
+                request.query().get_or("foo", "This is only a default value!"))
     });
 
     // try calling http://localhost:6767/strict?state=valid
     // then try calling http://localhost:6767/strict?state=invalid
     router.get("/strict", middleware! { |request|
-        if &*request.query("state", "invalid")[0] != "valid" {
+        if &*request.query().get_or("state", "invalid")[0] != "valid" {
             (BadRequest, "Error Parsing JSON")
         } else {
             (StatusCode::Ok, "Congratulations on conforming!")
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -84,14 +84,17 @@ fn main() {
 
     // try calling http://localhost:6767/query?foo=bar
     router.get("/query", middleware! { |request|
-        format!("Your foo values in the query string are: {:?}",
-                request.query().get_or("foo", "This is only a default value!"))
+        if let Some(vals) = request.query().all("foo") {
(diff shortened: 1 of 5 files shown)


cburgdorf:
Looks good to me :+1:

Ryman:
@cburgdorf The current code or the suggestions in the description?

cburgdorf:
Sorry, for the late reply. I'm traveling through southeast asia and only work sporadically. I meant your suggestion sounds good. And also, I'm more in favor of returning an empty slice instead of `None`.

Ryman:
@cburgdorf No worries, enjoy your time!

I don't think returning an empty slice is a good idea although we currently do it in `param` whenever "format" is used, but I want to fix that soon! `QueryStrings` and `Path`s are slightly different in this manner though, as the `param` in a path should never be empty (I wonder if there's a test for that :P) while a querystring having a blank value is legit.

Consider wanting to update a value to empty, "../update?bar=&car=hello". The user's server code will have no way to know if `bar` is set and empty as opposed to being unspecified, unless they parse the url themselves. It ends up with edgecases where [this](http://stackoverflow.com/questions/400364/asp-net-empty-querystring-parameter/3596770#3596770) or [this](http://stackoverflow.com/questions/4382669/checking-for-null-and-missing-query-string-parameters-in-php/4382690#4382690) seem reasonable.

cburgdorf:
Ok, true. Didn't think about such cases.

Ryman:
@cburgdorf Updated

daogangtang:
r +1

Ryman:
Any final opinions, or should this merge?

cc @cburgdorf @SimonPersson @rgs1 @ninjabear @lambdaburrito <anyone else?>

SimonPersson:
I'm terribly sorry for note helping out lately. I'm writing my thesis so there's no time. I'll be more helpful in a month or two!

SimonPersson:
From my quick look, I think we should merge this.

cburgdorf:
Ok, I went ahead and merged it so that we can get it into the new release that I'm about to publish on crates.io soon.

PR #196, opened 2015-04-19
Improve the error message on failure to parse JSON

jimtla:
rust-lang/rust#23979 has landed, so we can have nice JSON error messages!


jimtla:
Is `json` a reasonable scope? The doc doesn't give much description on what makes sense there.

cburgdorf:
Yep, not too strict about that ;)

Issue #136, opened 2015-01-25
error: the trait `nickel::middleware::Middleware` is not implemented for the type `fn(&nickel::request::Request<'_, '_>, &mut nickel::response::Response<'_, '_>)


tarcieri:
It seems that many old programs and examples do not work because of this?

I am trying to use @rgs1's `more-fixes-for-alpha` branch on 1.0.0-alpha

SimonPersson:
Yes, we believe this to be some sort of bug in the compiler. You can see the workaround in the examples in his fork, basically you just have to store the handler function in a variable before you add it to the router. Alternatively you can use the router macro instead.

We can keep this issue open so that we remember to look into it, hopefully it will just start working again one day.

Ryman:
@tarcieri It's a coercion failure, something about all `fn`s are now unique items and not function pointers unless cast to one. See https://github.com/rust-lang/rust/issues/20178 for more meaningful info.

In this case, if you have a handler `foo`, you can use something like `bar.something(foo as fn(&Request, &mut Response) -> MiddlewareResult);`

rgs1:
Hi @SimonPersson, @Ryman:

So Hyper does something like this to work around the need for casting (rust issue #20178):

```rust
trait Foo {
    fn foo(&self);
}

impl<F> Foo for F where F: Fn() {
    fn foo(&self) {
        self();
        println!("do foo");
    }
}

fn bar() {
    println!("do bar");
}

fn take_foo<F: Foo>(f: F) {
    f.foo();
}

fn main() {
    take_foo(bar);
}
```
This would allow us to work around the need for casting. The only issue that I found when trying to implement this is that when doing it, i.e., for:

```rust
impl<F, R> Middleware for R
    where F: Fn(&Request, &mut Response) -> R, F: Sync + Send + 'static, R: ResponseFinalizer {
    // ....
}
```

it mostly works except that the compiler then complains about conflicting implementations when doing it for:

```rust
impl<F, R> Middleware for R
    where F: Fn(&mut Request, &mut Response) -> R, F: Sync + Send + 'static, R: ResponseFinalizer {
    // ....
}
```
which is, allegedly, almost the same except for the Request being mut. If we could work around this, this might help us clean up the casts (I think!).

Thoughts?

Ryman:
[...]

PR #197, opened 2015-04-23
feat(unboxed_closures): allow unboxed closures for handlers and error handlers

Ryman:
This is a bit of a compromise, allowing unboxed closures to be used for handlers but not for `Middleware`, while also making the routing a bit less flexible for the moment (can only take unboxed closures and not a `Middleware`, so you can't have a `Router` dispatch to another `Middleware` without wrapping a closure).

The second commit shows an alternative that I encountered while refactoring. If we deem macros acceptable in the time being we could follow that approach, as we maintain flexibility with only a minor syntactic wart. I favor this approach, but wanted to show how the example usage could look without any macros, which is the current state of the PR. Whenever https://github.com/rust-lang/rust/issues/24680 is fixed, we should then be able to depreciate the macro and the user code should be very easy to update `middleware! { |req, res| "foo" }` becomes `|_, res| res.send("foo")`.

Another alternative is to use `Fn(..)` as the bound everywhere in future when implementing unboxed closures for a type becomes stable, then the `Middleware` trait will become redundant and we could implement `Fn(..)` for `Router` and such. This approach doesn't help us with the current 1.0 feature-set though.

### DIFF
--- a/examples/example_route_data.rs
+++ b/examples/example_route_data.rs
@@ -1,25 +1,13 @@
-extern crate nickel;
+#[macro_use] extern crate nickel;
 
-use nickel::{Nickel, Request, Response, HttpRouter, MiddlewareResult};
+use nickel::{Nickel, HttpRouter};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::Relaxed;
 
-struct Logger {
-    visits: AtomicUsize
-}
-
 fn main() {
     let mut server = Nickel::new();
+    let visits = AtomicUsize::new(0);
 
-    fn root_handler<'a>(_: &mut Request, response: Response<'a>, logger: &Logger)
-            -> MiddlewareResult<'a> {
-        let text = format!("{}", logger.visits.fetch_add(1, Relaxed));
-        response.send(text)
-    }
-
-    // issue #20178
-    let rhandler: for <'a> fn(&mut Request, Response<'a>, &Logger) -> MiddlewareResult<'a> = root_handler;
-
-    server.get("/", (rhandler, Logger{visits: AtomicUsize::new(0)}));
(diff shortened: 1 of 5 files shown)


Ryman:
Some example usage:

Macro-less version where router is bounded by `Fn` instead of `Middleware`
```rust
let visits = AtomicUsize::new(0);
router.get("/", move |_, response|
    response.send(format!("{}", visits.fetch_add(1, Relaxed)))
);

// but unable to do something such as:
router.get("/other", SomeMiddleware);
```

Retaining the macros which now can use an unboxed closure under the hood, everything sticks with `Middleware` bounds.
```rust
let visits = AtomicUsize::new(0);
router.get("/", middleware! { |_req, _res| format!("{}", visits.fetch_add(1, Relaxed)) });

// can also still do this
router.get("/other", SomeMiddleware);
```

cburgdorf:
I think a router that doesn't accept a `Middleware` anymore isn't a good compromise for now.

However, your other attempt looks pretty cool.

```rust
let visits = AtomicUsize::new(0);
router.get("/", middleware! { |_req, _res| format!("{}", visits.fetch_add(1, Relaxed)) });

// can also still do this
router.get("/other", SomeMiddleware);
```

This seems to be a good compromise as it allows closure usage with just a little syntactic bloat (the macro). And as rust gets better over time one will just be able to drop the macro and keep the closure as it is. :+1:

bodokaiser:
Hi,

I still get the error from #136 when trying to compile [these two files](https://github.com/bodokaiser/rust-restful/blob/master/src/main.rs#L19-L21).

Any ideas?

Ryman:
@bodokaiser Opened https://github.com/bodokaiser/rust-restful/pull/1 with a fix.

Issue #188, opened 2015-04-13
Doesn't keep the server alive


cburgdorf:
This doesn't keep the server running.

```rust
#[macro_use] extern crate nickel;

use nickel::Nickel;
use nickel::router::http_router::HttpRouter;

fn main() {
    let mut server = Nickel::new();

    server.get("/bar", middleware!("foo"));
    server.listen("127.0.0.1.6767");
}
```

Ryman:
Should be `127.0.0.1:6767` ?

cburgdorf:
Oooops, how is that not a compile time error?

Ryman:
It should be a runtime error? http://is.gd/ynNsUe

Ryman:
Fixed in #198

PR #198, opened 2015-04-26
fix(server): ensure panic on invalid listen addr

flaub:
Related to issue #188.


cburgdorf:
Looks good to me. What does @Ryman say?

Ryman:
Looks good, but I would prefer if the test had some more reasonable cases like the one in the other issue "127.0.0.1.6767", and if you want to add a doc that it may panic it would be neat but not necessary for now!

Ryman:
Merged as https://github.com/nickel-org/nickel.rs/commit/e218287c96064ee7af4e5f899ae186b048696048, Thanks!

PR #199, opened 2015-04-27
Allow middleware to bind response mutably

VShell:
If they can't bind response mutably, they can't call &mut methods on the Response and set headers. Otherwise, to use `response.content_type()`, middleware creators must create a function instead of using the middleware! macro.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -7,7 +7,7 @@ use std::io::Write;
 use nickel::status::StatusCode::{self, NotFound, BadRequest};
 use nickel::{
     Nickel, NickelError, Continue, Halt, Request,
-    QueryString, JsonBody, StaticFilesHandler, HttpRouter, Action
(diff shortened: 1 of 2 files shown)


Ryman:
Can you add an example of this usage? (or update one of the examples?)

Ryman:
@VShell Thanks!

(Also, just in case you weren't aware, there is no notification when you push new commits to a PR, only for comments and such)

ninjabear:
This is well handy - could the same thing be applied to the ```router!``` macro? 

e.g.

```rust
   nickel.utilize(router!(

     get "**" => |request, mut response| {

        let mut data = HashMap::<&str, &str>::new();
        data.insert("title", "hello world");

        response.render("templates/my_template.mustache", &data);
     }

   ));
```

Ryman:
@ninjabear Should be, I've been meaning to rewrite it recursively for other reasons and I think it would be a requirement for the above to work. I guess I should prioritize that!

PR #200, opened 2015-04-28
add https interface.

daogangtang:
this PR contains the following two features:
1. add https interface. now you can use  server.listen_https(addr, cert, key) to run a https server;
2. make urlencoded mod public.


src/lib.rs
```diff
@@ -55,7 +55,7 @@ mod static_files_handler;
 mod json_body_parser;
 pub mod mimes;
 mod query_string;
-mod urlencoded;
+pub mod urlencoded;
```
Ryman: A little wary on publicly exposing this, what is the requirement? (and can it be separated from this PR)

src/nickel.rs
```diff
@@ -141,4 +142,15 @@ impl Nickel {
...
+    
+    pub fn listen_https<T: ToSocketAddrs + Display>(mut self, addr: T, cert: &Path, key: &Path) {
+        self.middleware_stack.add_middleware(middleware! {
+            (StatusCode::NotFound, "File Not Found")
+        });
+
+        println!("Listening on https://{}", addr);
+        println!("Ctrl-C to shutdown server");
+
+        Server::new(self.middleware_stack).serve_https(addr, cert, key);
```
Ryman: I think `Server::new` should take an enum describing if it's HTTP or HTTPS, and then `Nickel::listen_http` defined in terms of `Nickel::listen_https` or at least factoring out common code. I don't think we need to add additional api to `Server` since it's internal and therefore not a big deal to have to import an enum variant.

Ryman: Also, I think client API would benefit from taking `AsRef<Path>` for cert and key, what do you think?

daogangtang: maybe your scheme is better.

Ryman:
@daogangtang Any update on this?

Ryman:
Closing due to inactivity, feel free to open a new PR if you want to take another try at this.

PR #206, opened 2015-05-01
Fix instructions about nightly in readme

tshakah:
Nickel now builds and runs fine on nightly without the flag (in fact the flag gives an error) - I presume this is something to do with https://github.com/nickel-org/nickel.rs/pull/205?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,13 +52,7 @@ Then try `localhost:6767/user/4711` and `localhost:6767/bar`
 
 ### Note about nightly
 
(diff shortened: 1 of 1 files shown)


Ryman:
Ah, yes, I should have updated the readme to `unstable` instead of `nightly`. I'd prefer to change to that instead of saying it doesn't require it, as it may again require it in the future so it seems more proof to leave it and just fix the error?

Ryman:
Closing via #216 and #220

PR #210, opened 2015-05-06
fix(server): loosen bounds required for `listen`

### LINKED ISSUE #208: Display trait in Nickel::listen prevents using (addr, port) tuple
Nickel::listen expects a ToSocketAddrs, which is conveniently implemented by (&str, u16). This allows to do something like server.listen(("localhost", port))

However, requiring Display prevents this tuple from being used there.

Is it possible to drop the Display requirement? As the compiler suggests, "try using `:?` instead if you are using a format string".

(I tried implementing Display for the tuple, but apparently I can't do that from my crate.)

As a workaround, I am currently using

    server.listen(net::SocketAddrV4::new(net::Ipv4Addr::from_str("localhost").unwrap(), port));

But it's not quite as nice...

Ryman:
This also pins the hyper dependency as there's some upcoming breakage coming in 0.4 that we'll need to adjust for.

Fixes #208.


Ryman:
Going ahead to merge as this should also fix #212

PR #213, opened 2015-05-08
Try to keep the README examples more in sync with changes

Ryman:


Ryman:
@cburgdorf r?

cburgdorf:
Good move ;)

cburgdorf:
Ok, I was quick to hit the merge button. But actually why not make it even simpler.

```rust
#[macro_use]
extern crate nickel;

use nickel::Nickel;
use nickel::router::http_router::HttpRouter;

fn main() {
    let mut server = Nickel::new();
    server.get("/hello", middleware!("hello world"));
    server.listen("127.0.0.1:6767");
}
```

Ryman:
@cburgdorf I'm not personally fussed about what the example is, as long as it's buildable :) Feel free to update it :+1: 

The main reason I changed it is that to get an example with `extern`s other than nickel working it required putting their hash into the rustdoc command, which would be overly brittle :(

PR #214, opened 2015-05-08
feat(router): add convenience methods for OPTIONS and PATCH

### LINKED ISSUE #202: Add support for OPTION requests
I currently build something on top of nickel where I need to support CORS. To support CORS we need to allow OPTION requests. Maybe we can do even more to make CORS easier with nickel.

/cc @pascalprecht

Ryman:
Closes #202


cburgdorf:
:+1:

PR #215, opened 2015-05-10
lambdaburrito's PR tidied up

ninjabear:
@lambdaburrito raised a PR here (https://github.com/nickel-org/nickel.rs/pull/190) but I think he may be MIA now. It's only a minor change so I did the tidy up for him/her. 

I also renamed the variables from ```regex1``` etc and made them a bit more descriptive


Ryman:
Minor nits if you feel you want to add more:
- the tests would be better if they assert the data is what's expected
- `assert_eq!(.., true)` would give a better error message if it was just `assert!(..)`
- `regex_x`, `regex_y`, `regex_z` could probably just be `x`, `y`, `z`

If not, I'm quite happy to merge as is and I can add to it later :)

ninjabear:
like this @Ryman ?

Ryman:
Thanks :+1:

Issue #159, opened 2015-03-05
Documentation doesn't list what to put in cargo.toml


ninjabear:
Not sure if this is anywhere already but if it isn't it would be helpful.

```{toml}
[dependencies.nickel]

git = "https://github.com/nickel-org/nickel.rs.git"

[dependencies.nickel_macros]

git = "https://github.com/nickel-org/nickel.rs.git"
path = "nickel_macros"
```

I came out with this for now but no idea if it's "correct"

Ryman:
The `path` on the second import shouldn't be necessary, cargo does a search throughout the whole repo for Cargo.toml's with libs defined :)

Additional note: we can move nickel_macros into nickel now since we probably aren't going to steer towards compiler plugins as they're not a 1.0 feature.

Ryman:
I think this is covered now :)

PR #216, opened 2015-05-10
Reword the readme and split contributor information into another file

ninjabear:
#159 was the original source of this, but I remembered finding it quite hard to understand the readme before because it contains a lot of information on various aspects of nickel development, imo all it really needs to have is a status, how to helloworld and where to find out more. 

I appreciate the readme is pretty subjective and am not suggesting this PR be accepted entirely as is - I'm open to changing it up

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,64 +1,31 @@
-nickel.rs
-=======
-[![Build Status](https://travis-ci.org/nickel-org/nickel.rs.png?branch=master)](https://travis-ci.org/nickel-org/nickel.rs)
-[![](http://meritbadge.herokuapp.com/nickel)](https://crates.io/crates/nickel)
-[![Join the chat at https://gitter.im/nickel-org/nickel.rs](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/nickel-org/nickel.rs?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+# [nickel.rs](http://nickel.rs) [![Build Status](https://travis-ci.org/nickel-org/nickel.rs.png?branch=master)](https://travis-ci.org/nickel-org/nickel.rs) [![](http://meritbadge.herokuapp.com/nickel)](https://crates.io/crates/nickel) [![Join the chat at https://gitter.im/nickel-org/nickel.rs](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/nickel-org/nickel.rs?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,64 +1,31 @@
...
 
-To build on nightly, you will need to add `--features nightly` to the end of the above commands. Alternatively, if depending on the library you will need to add the following to your `Cargo.toml`.
+name = "my-nickel-app"
+version = "0.0.1"
+authors = ["yourname"]
 
-```not_rust
 [dependencies.nickel]
 version = "*"
 features = ["nightly"]
```
Ryman: This is only required for nightly, and it's "unstable" now instead of "nightly" :sob:

README.md
```diff
@@ -1,64 +1,31 @@
...
+
+[dependencies]
+rustc-serialize = "*"
```
Ryman: `rustc-serialize` was required for the old readme example before I trimmed it down, it's now only needed for some of the examples. It's probably worth adding some sort of explanation within the examples themselves (possibly comments beside the imports) or some comment similar to 'you can find the crates required to build the examples on crates.io'. Not sure which is best!

README.md
```diff
@@ -78,20 +45,22 @@ fn main() {
...
 
-##License
+## More examples
+
+More examples can be found [in the examples directory](/examples/) and the full documentation can be [found here](http://nickel-org.github.io/nickel/).
 
-Nickel is open source and licensed with the [MIT license](https://github.com/nickel-org/nickel/blob/master/LICENSE)
+##License
 
+[nickel.rs](http://nickel.rs) is open source and licensed with the [MIT license](https://github.com/nickel-org/nickel/blob/master/LICENSE)
```
Ryman: Can probably drop most of the line pre- "[MIT](..)"

Ryman:
This looks like a good improvement :+1: 

I've been meaning to get round to this for a while, my preference would be to cut it down even a bit more (I'm kind of [inspired by this recently](https://speakerdeck.com/rowanmanning/ux-for-your-node-modules)). I do think we should slip the example code above the installation instructions for sure!

ninjabear:
I really like the readme proposed in that link too - its kind of where I was going with this but I think I've missed some bits out and given some things slightly odd names (Usage->hello world). I can change it to be more like that if you like @Ryman ?

Ryman:
My preference would be to put the code sample where dependencies are right now, without any titles (its implied) and then put dependencies below as 'Installation' or similar. But other than that I think most of this PR is worth taking as-is to iterate on :+1: 

@cburgdorf any opinion on this?

ninjabear:
@Ryman I went through your comments and implemented, think I caught them all? 

I changed "nightly" to "unstable" but I'm not sure if you think we should remove the channel altogether?

Ryman:
Looks good, much appreciated @ninjabear!

ninjabear:
np :)

PR #217, opened 2015-05-10
WIP: address a number of ergonomic issues

Ryman:
There's a number of breaking changes in here, each commit outlining the extent of breaks.

The majority of the changes are focused on improving the api for `Response` and removing access to it's implementation details. The main change is adding a dependency on [`modifier`](https://github.com/reem/rust-modifier) allowing us to make most interactions with `Response` through `Response::set`, with the explicit `headers`/`status` available for less common cases.

`router!` has been changed to allow more flexible parameters per route similar to the `middleware!`, so I've now enabled the warnings for unused parameters again to match up with expectations of rust code.

Would like to have some feedback on this, I'm not fond of the `nickel_try` macro, I'm sure there's something better around that idea, but I felt it was worth putting it in for discussion.


Ryman:
Updated `nickel_try!` to `try_with!` which I'm much happier about.

Ryman:
Here's an implementation of todo-backend using this branch: https://github.com/Ryman/nickel-todo-backend

cburgdorf:
That's pretty cool. It's also a good starting point to move the nickel.rs website to heroku, I think :)

I also looked through each commit and it all looked sane to me. Event though I probably can't follow everything to the last bit ;)

While we are at ergonomics. I noticed Iron supports closures without the need of a macro. 

```rust
    Iron::new(|req: &mut Request| {
        Ok(Response::with((status::Ok, "Hello World")))
    }).http("localhost:3000").unwrap();
```

What's the reason they don't run into the same type inference restrictions that we do?

Ryman:
@cburgdorf Our handlers link a lifetime from an input position to an output, which seems to be the problematic part, as you can't typehint lifetimes well enough currently (afaik) without using the hacks in the macros that we have. I guess Iron doesn't have the lifetime linking issue as they create the Response within their handler.

I've added some compile-fail checks so that we can get a fail when one of the examples starts to pass (then we know we should look into trying again!). Also, I had to disable benchmarking on travis as the hyper/nickel build was taking 10minutes with `opt-level=3` which causes travis to kill the build.

Ryman:
@lambdaburrito This was the PR I mentioned yesterday for feedback!

lambdaburrito:
@Ryman There's too many changes to comment on individually but I think this a really great step to evolve the API :+1: Great work! I really like what you have done with `router!` code as it will clean up my code. By the way, it was cool to put a face to a name on Friday :)

cburgdorf:
>it was cool to put a face to a name on Friday :)

Now I'm jealous.

Ryman:
@lambdaburrito Same, hope the meetups continue semi-regularly this time! Thanks for the feedback :)
@cburgdorf You'll reach London sometime I'm sure! :wink: 

So, there hasn't really been any issues raised, so I'll look to merge this tomorrow if nothing else pops up?

Ryman:
Merged and published new version

PR #228, opened 2015-05-25
Publish docs on build

Ryman:
Hopefully got it setup correctly!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,12 @@ script:
       travis-cargo doc &&
       echo "Testing README" &&
       rustdoc --test README.md -L dependency=./target/debug/deps --extern nickel=./target/debug/libnickel.rlib
+
+after_success:
+  - |
+      (echo "docs.nickel.rs" > target/doc/CNAME) &&
+      travis-cargo --only 1.0.0 doc-upload
+
+env:
+  global:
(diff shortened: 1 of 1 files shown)


Ryman:
@cburgdorf 
Okay, it half worked. 
The current behaviour:
- http://docs.nickel.rs/ : 404
- http://docs.nickel.rs/nickel : old docs from nickel-org.github.io repo
- http://docs.nickel.rs/nickel.rs : new docs

Currently, the CNAME from https://github.com/nickel-org/nickel-org.github.io seems to be overriding this repo, so docs.nickel.rs 404's (due to no index in that repo), rather than aliasing nickel-org.github.io/nickel.rs as it should.

As far as I understand, we should move the `site` repo, to the `nickel-org.github.io` repo with the CNAME being 'nickel.rs', and then 'docs.nickel.rs' will be unique to this repo and will alias http://nickel-org.github.io/nickel.rs which works as expected (new docs). 

I'm guessing I can maintain repo history for `site` by cloning, adding a new remote to `nickel-org.github.io`'s repo and then force pushing to `gh-pages`? Would you be happy enough with these changes?

Ryman:
Another option is to just delete the `nickel-org.github.io` repo and rename `site` which would benefit from github's redirection for renames (no need to update git remotes).

cburgdorf:
I think either way is fine but I'd probably try to just delete the `nickel-org.github.io` and then rename `site`. No worries from my side :)

Ryman:
Deleted and renamed, will make a minor change to nickel.rs (old 'site') in a bit to trigger a rebuilt to know if the rename went OK for that one at least.

Ryman:
[nickel.rs](http://nickel.rs) && [docs.nickel.rs](http://docs.nickel.rs) are now working! :sweat_smile:

Issue #229, opened 2015-05-31
RouteResult's `param` function is inflexible


dallbee:
From here, in router.rs: 
```rust
impl<'a> RouteResult<'a> {
    pub fn param(&self, key: &str) -> &str {
        for &(ref k, ref v) in &self.params {
            if k == &key {
                return &v[..]
            }
        }

        // FIXME: should have a default format
        if key == FORMAT_PARAM { return "" }
        panic!("unknown param: {}", key)
    }
}
```

A panic is returned, not allowing for things like 
```rust
pub fn handler<'a>(req: &mut Request, res: Response<'a>) -> MiddlewareResult<'a> {
    // check for existence of req.param("foo") and act on it
}

server.get("/", handler);
server.get("/:foo", handler);
```

This could be done by changing param to use an Option, or even simply changing the error mode of param to be failure instead of panic, allowing for handling of the error.

Ryman:
Closing as #232 merged. Please reopen if you still have issues!

dallbee:
Neat. Thanks!

PR #232, opened 2015-06-09
fix(request): request.param(_) now returns an Option

Ryman:
A user may wish to use the same handler for multiple routes,
if the behaviour of such a handler should differ based on the
available params then the handler it will panic at runtime as
there is no convenient way to check for the presence of the
param before accessing it.

For broken code:
If you have `foo.param("bar")` then you'll need to `unwrap` the
Option, e.g. `foo.param("bar").unwrap()`.

[breaking-change]

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -43,7 +43,7 @@ fn main() {
 
     // go to http://localhost:6767/user/4711 to see this route in action
     router.get("/user/:userid", middleware! { |request|
-        format!("This is user: {}", request.param("userid"))
+        format!("This is user: {}", request.param("userid").unwrap())
     });
 
     // go to http://localhost:6767/bar to see this route in action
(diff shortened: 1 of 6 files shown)


Ryman:
This would cover #229 and I wanted to push up what a naive implementation would do, I'm not really sure I like it as it's a lot of verbose unwraps which are *known* to be valid in many cases. But on the other hand, adding another `opt_param` accessor is leaving the current behavior of unexpected runtime panic for the user using `param` (although a warning in the docs could be *good enough* as there is precedent in the stdlib?). 

I have some thoughts on some macro workarounds to get rid of verbose unwraps, but having the non-macro case be usable is important (since shared handlers are usually defined as functions).

*Also*, I'm allowing failures on nightly travis as stable contributions shouldn't have to concern themselves too much with that.

cc @cburgdorf @dallbee

Ryman:
@SimonPersson @cburgdorf Rebased on master, can we get some opinions on this before merging a version bump from #237?

SimonPersson:
This is great. Obviously an improvement to the current situation. I say merge it! :+1:

cburgdorf:
Yep, looks good to go!
Am 12.07.2015 16:24 schrieb "Simon Persson" <notifications@github.com>:

PR #233, opened 2015-06-24
fix(responder): relax required lifetime for responding with string slices

Ryman:
@cburgdorf This is related to the issue you mentioned on gitter


cburgdorf:
Thanks for the fix. Can I just borrow your brain for a bit ;)

Issue #98, opened 2014-10-11
Provide authentication & session middleware


jacek99:
It would be great if the default middleware included an out of the box solution for authentication and session handling.

That seems to be the most missing part in order to get an actual production grade app written in Nickel

dariusc93:
Any updates on implementing sessions into nickel?

cydrobolt:
Is this implemented yet?

cburgdorf:
@cydrobolt we have https://github.com/nickel-org/session and https://github.com/nickel-org/cookies and I'm using them in production. 

One thing to note is that they are not on cargo so you have to do the following

```
[dependencies.nickel_cookies]
git = "https://github.com/nickel-org/cookies.git"

[dependencies.nickel_session]
git = "https://github.com/nickel-org/session.git"
```

We don't have something polished for Auth so yet.

s12me:
Is Auth available now ?

jolhoeft:
Auth is still somewhat a roll your own situation yet. I have a crate, https://crates.io/crates/websession, which I use for authentication. The documentation for how to use it with nickel isn't really there yet, though. I am inclined to think Auth ought to be a separate crate from nickel.

PR #234, opened 2015-06-27
feat(cookies): Adds support for cookies.

SimonPersson:
Related to #98.

Makes it possible to set and read cookies.


Ryman:
I don't really like the API, could we perhaps have 2 traits `Cookies` and `CookiesMut` where both involve a `cookie::CookieJar`? The `CookieJar` seems to be quite flexible and means the user won't have to worry so much about the finer details of how cookies work.

I imagine the request would only implement `Cookies`, while the response could do both `Cookies` and `CookiesMut` (providing respective `cookies` and `cookies_mut` methods?). Both would probably require using a plugin as parsing the `CookieJar` for the `Request` would need to be stored, and the `Response` would need to store it somewhere. 

I guess there is the consideration around "when do the cookies get applied to the response" if using a plugin, I wonder if that's a gap in the plugin api and there needs to be some kind of `on_send` event where the `CookieJar` could be serialise into the `Response` itself, via [`SetCookie::from_cookie_jar`](http://hyper.rs/hyper/hyper/header/struct.SetCookie.html#method.from_cookie_jar) when the headers are being sent.

What do you think? Should I sketch it out perhaps or do you want to give it a go?

I'll comment more on the auth PR with some related thoughts.

SimonPersson:
@Ryman - Go for it! I only threw something together so that I could work on auth. It sounds like you have some good ideas :)

Ryman:
Closing in favor of #272

Issue #79, opened 2014-09-27
Type  safety for Request/Middleware


SimonPersson:
The following code builds today:

    extern crate nickel_postgres;
    extern crate nickel;

    use std::io::net::ip::Ipv4Addr;
    use self::nickel::{Nickel, Request, Response, JsonBody};
    use nickel_postgres::{ PostgresMiddleware, PostgresRequestExtensions };

    fn main() {
        let mut server = Nickel::new();
        fn db_handler(req: &Request, response: &mut Response)
        {
            let db_conn = req.db_conn();
            response.send("doesn't work");
        }
        let mut router = Nickel::router();
        router.get("/", db_handler);

        server.utilize(router);
        server.listen(Ipv4Addr(127, 0, 0, 1), 6767);
    }

`db_handler()` will never be able to retrieve a database connection since we haven't utilized a PostgresMiddleware and we will get a crash at every request.

In my opinion, the whole point of writing web applications in a statically typed language is that the compiler will catch this kind of bugs for us. With the current situation we're no better than Python/Ruby/Whatever in that regard.

Could we discuss some approaches to solving this? I have some ideas that I'm going to try out but I'd like to hear some more thoughts on this too.

cburgdorf:
Closing as part of a clean up of our issue tracker. Anyone please free to reopen for further discussion.

PR #235, opened 2015-07-04
Cookies #2

Ryman:
cc @SimonPersson 

This took a bit more thinking than I'd anticipated, but it required something similar to solving #79 for some configuration without incurring ordering dependencies (which could be potentially insecure for signed cookies) that seems slightly obvious in hindsight (and is very similar to something you suggested there). Still feels a bit iffy though, but I think it's sound enough to be useful.

The last commit does a bit of weirdness with where clauses on `Self` to collapse the `Cookies` trait into a single trait, but I think it's a lot nicer for users to only have to worry about importing a single trait.

cc #234 (original cookies implementation)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ groupable = "*"
 mustache = "*"
 lazy_static = "*"
 modifier = "*"
+cookie = "*"
 
 [dependencies.compiletest_rs]
 version = "*"
@@ -37,6 +38,11 @@ optional = true
 
 name = "example"
 path = "examples/example.rs"
+
+[[example]]
+
+name = "cookies_example"
+path = "examples/cookies_example.rs"
 
 [[example]]
 
--- a/examples/cookies_example.rs
+++ b/examples/cookies_example.rs
@@ -1,0 +1,28 @@
+#[macro_use] extern crate nickel;
+extern crate cookie;
+
+use nickel::{Nickel, HttpRouter, Cookies};
+use nickel::cookies;
+
+struct Data {
+    secret_key: cookies::SecretKey
+}
+
+impl AsRef<cookies::SecretKey> for Data {
+    fn as_ref(&self) -> &cookies::SecretKey {
+        &self.secret_key
+    }
+}
+
+fn main() {
+    let data = Data { secret_key: cookies::SecretKey([0; 32]) };
+    let mut server = Nickel::with_data(data);
+
+    // Try curl -b MyCookie=bar localhost:6767
+    server.get("/", middleware! { |req|
+        let cookie = req.cookies().find("MyCookie");
+        format!("MyCookie={:?}", cookie.map(|c| c.value))
+    });
+
+    server.listen("127.0.0.1:6767");
+}
--- a/examples/cookies_example.rs
+++ b/examples/cookies_example.rs
@@ -1,7 +1,7 @@
 #[macro_use] extern crate nickel;
 extern crate cookie;
 
-use nickel::{Nickel, HttpRouter, Cookies, CookiesMut, QueryString};
+use nickel::{Nickel, HttpRouter, Cookies, QueryString};
 use nickel::cookies;
 use cookie::Cookie;
 
--- a/examples/cookies_example.rs
+++ b/examples/cookies_example.rs
@@ -1,8 +1,9 @@
 #[macro_use] extern crate nickel;
 extern crate cookie;
 
-use nickel::{Nickel, HttpRouter, Cookies};
+use nickel::{Nickel, HttpRouter, Cookies, CookiesMut, QueryString};
 use nickel::cookies;
+use cookie::Cookie;
 
 struct Data {
     secret_key: cookies::SecretKey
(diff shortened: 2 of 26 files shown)


SimonPersson:
Cool!

I took a quick glance, and the new API looks nice to me. :)

Is this a breaking change?

Ryman:
Ah, yes forgot about that, I'll annotate the commits when I get a chance later (as well as test a beta build which seems to have broken from inference changes in 1.2?)

Ryman:
Managed to get a build on 1.1 and beta (1.2)... which ICEs on 1.0. Sweet.

Ryman:
Alright, things work across rust versions and have an additional compile-fail test to check that the error message is 'reasonable' when using cookies without the appropriate bounds being met.

cc @cburgdorf

cburgdorf:
Sorry, for jumping on it late. I skimmed through the changes and it's quite a brain twister for me to follow ;)

Looking at [the example code](https://github.com/Ryman/nickel.rs/blob/2cdac29f757f5e25aead58b89234e249dee343a9/examples/cookies_example.rs#L19-L26) I wonder what's the advantage of having the user to define a data structure for the cookie. Iron just defines `Cookie` for you which seems to make it easier from the perspective of the user: https://github.com/iron/cookie/blob/60832b7dd3020724bf2a0cecb631f59acfa7b0e0/README.md

There may be good reasons to implement it our way but I'd like to understand them better ;)

SimonPersson:
That struct is needed to encrypt the Cookie, which Iron does not appear to be doing. So in Iron, the user would instead have to manually deal with encrypting whatever data they want to send, but as they don't have an example for that, Iron does come across are easier :)

My suggestion is that we provide a struct so that the user doesn't have to create one in the most simple case. Then the example could be:
```
let mut server = Nickel::with_data(SecretKey::new_with_rand_key());
```
or something, and the user would not have to create a struct and initialize it. We could also add a case for those who prefer use the same key:
```
let mut server = Nickel::with_data(SecretKey::new("abcdef123456789"));
```

What do you think, @Ryman?

Ryman:
That repo looks like it hasn't been updated in a year, but to be clear, the advantage is that this way will offer a compile-time guarantee, as you can see [this error message is given when not handled](https://github.com/Ryman/nickel.rs/commit/2cdac29f757f5e25aead58b89234e249dee343a9#diff-2654a9417339b2c33737f59cfa3dce98R19).

If we emulated the way a usual plugin would do it, then we get the dependency problem where you'll have a runtime crash due to the unwrap (see the first line in `echo_cookies` with the trailing unwrap) if the Cookies middleware is added after a handler which depends on it.

If we had specialization then I think we could provide a zero'd key for things which don't implement `AsRef<cookies::SecretKey>` and not allow the user to use `encrypted` or `signed` cookies, but I don't think there's a clean way to do that at present.

@SimonPersson It does/did handle encryption ([see here](https://github.com/iron/cookie/blob/master/src/parser.rs#L37)). But it does implement most of it by itself rather than leveraging cookie-rs (not sure if cookie-rs was available at the time though so it makes sense).

I think `Nickel::with_data(cookies::SecretKey([32,4,125,... ]))` works, but the problem is more general, like if there's two middleware which require bounds, e.g. cookies + database pool. The example code is perhaps trying to be too general by showing that the user is in control of the data and they're not restricted to `nickel` types. 

I don't think generating a new key on every server restart is a good idea as that invalidates all old cookies, providing the `SecretKey::new` could be useful though. 

Perhaps we need to consider how configuration will look for nickel in general, to place my vote, I'm in favor of sane implicit defaults (even so far as the server could autogenerate a config on first boot if no config file is available). This would put us nowhere near 'express like' though and align more with something like rails.

SimonPersson:
@Ryman: It only signs the cookie, no encryption, no? Or am I missing something?

Perhaps we can break the example out to two examples, one where we show the most simple use of Cookies and one where we demonstrate sharing of data. I suppose the `nickel-auth` crate will also serve as a good example of this once it is done.

I suppose I prefer that users who know nothing about encryption gets a safe key generated, and having the cookies become invalidated on a server restart, then having the same user provide an insufficiently strong key. Users who know what they're doing can still provide their own key. I very much like the idea of generating a config file though! Then we would get around that problem entirely by storing a generated key for the user.

Ryman:
@SimonPersson Ah yes I'm totally wrong about the encryption :P

In the simple use of cookies, do you mean with no key, or with just passing something like `SecretKey::new("Keystring")`?

Yeah, I would prefer a system where there are safe defaults rather than defaulting to something with a null key or no encryption. Perhaps we can elaborate on another ticket how we will handle simpler configuration, or do you think it's a blocking issue for cookies for now?

SimonPersson:
@Ryman To be fair though, it is a bit odd to sign but not encrypt, I was also a bit confused at first.

Yes, just passing a `SecretKey` without creating a new struct, so that no one thinks that it is necessary to create a struct in order to use cookies.

I don't think it is blocking cookies. We can always add that later. Feel free to open another ticket.

Ryman:
Closing in favor of #272

PR #237, opened 2015-07-10
Update hyper to 0.6

Manishearth:


Manishearth:
(note that without this nickel won't compile on latest nightly due to some modified lints which are set to deny in hyper.)

Manishearth:
@Ryman r?

Ryman:
@Manishearth Sorry for the delay, thanks for looking forward!

PR #238, opened 2015-07-12
chore(travis): Build on latest stable rust.

SimonPersson:
See title.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ cache:
       - target
 
 rust:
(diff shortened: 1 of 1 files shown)


Ryman:
You could use `stable` instead of a specific version, perhaps we can add both?

SimonPersson:
Yeah, let's keep 1.0.0 and add stable. Once we get a new stable, we can add 1.1.0 and go on like that until we need some new feature for some reason.

PR #243, opened 2015-07-23
chore(readme): add license badge

netmml:


Ryman:
Not sure which I prefer, but if we're to adopt this change can you bring them all back on to the side of the heading? (It looks like the license and the gitter link have fallen below)

Also can you amend the commit message to 'chore(readme): add license badge'?

netmml:
@Ryman fixed the problem, I'm doing this all in github's web interface so renaming this pr should do the trick

Ryman:
Sadly it doesn't update the commit message itself :( 

I squashed your commits, renamed and pushed it as https://github.com/nickel-org/nickel.rs/commit/b6e1ff96a3b6ff2408fab4c23c9acfd5ea7aca58. Thanks!

PR #246, opened 2015-08-01
Handling Sessions with an 'Environment'

Ryman:
This builds upon the ideas from #235 to add an 'environment' to each request. See https://github.com/nickel-org/nickel-auth/pull/1 for a related discussion which lead to this proposed change. `Session` would be really awkward to use correctly without something along the lines of this.

cc @SimonPersson


SimonPersson:
Exciting!

Should we perhaps document the type hinting for the middleware macro somewhere? I think that one example that uses it will be very easy to miss if one runs into this problem.

Other than that, I have no hesitation. Good job! :+1:

Ryman:
Rebased on #247 and updated https://github.com/Ryman/nickel.rs/commit/edea3ab3069e13903de34ce00cd19d788b83f133 with a note about how to use the middleware! macro with a type hint. 

I'm thinking we should add an explicit trait requirement for `Cookies` rather than  `AsRef<SecretKey>` as then we can provide documentation on the trait itself, perhaps providing a better name and explaining what the purpose is. 

I'm sure a lot of this needs a bit more annotation, so let me know where else to add!

SimonPersson:
Nice.

Yes, perhaps an explicit trait could be a good idea. We should probably await comments from someone else. You and I have been working on this for some time, so things that appear to be clear to us might be confusing to others.

cburgdorf:
I'm currently building a real world project with nickel that needs Session handling. I figured that this branch may be the most up to date and wanted to give it a try. I'm running into `Package `cookie v0.1.20` does not have these features: `secure`` though. What do I need to do to get this to compile?

Ryman:
It was added in cookie 0.1.21 by the looks of it, have you tried `cargo update`?

cburgdorf:
Ooops, bloody noob mistake. Never mind ;)

examples/session_example.rs
```diff
@@ -0,0 +1,72 @@
...
+
+fn main() {
+    let mut server = Nickel::with_data(ServerData);
+
+    /* Anyone should be able to reach thist route. */
+    server.get("/", middleware! { |mut res|
+        format!("You are logged in as: {:?}\n", res.session())
+    });
+
+    server.post("/login", middleware!{|mut res|
```
cburgdorf: I may not see the forest for the tree here. How do I fully test this example? Of course I can send the right json to the running server using something like curl or postman and I get the "successfully logged in" text. But I'd like to test out the entire example (session cookie). So I thought, ok, let's just render a small `<form>` for the `/` route that can be used for the login. But that's not gonna work as non ajax submitted forms aren't send as JSON so I would have to change other parts of the example, too. 

So I wonder, did you actually test the entire example to see if the session cookie is created by the browser after successful login? If so, how did you test that out?

@Ryman

Ryman: See https://github.com/nickel-org/nickel-auth/pull/1, it has a test shell script. (but yes we should have a test written in rust!)

cburgdorf:
I played with this a bit more and noticed that in the `session_example` it directly creates a `__SESSION` cookie even before `/login` is called. Is this intended behaviour? Does it automatically create a session because of the `Nickel::with_data(ServerData);` call where `ServerData` implements `SessionStore`?

How would I go if I would like to create the session just after a specific route was called?

Ryman:
@cburgdorf The session is quite eager, we could hold off on creating it until there is data written to it (as opposed to only read). In the example, when hitting '/' or '/secret', it would create the session as it gets read, fails to find an active session, and so creates a default one (`None` in this case). 

And yeah, you can't use session at all without implementing the `SessionStore` for your datatype (and there are a number of dependent traits, the store must be encodable, have a default, etc).

Can you shed light on why you want the Session cookie to only exist after a specific route is called?

cburgdorf:
I don't really need it that way but it was unexpected at least. Also some
people may be concerned of unnecessary request bloat. Think of a huge
website where only a minority (e.g admins) will ever make use of the
session. It will create a session cookie for everyone else even if it's not
used/needed.
Am 10.08.2015 03:54 schrieb "Ryman" <notifications@github.com>:

Ryman:
It's a fair point, so perhaps we should only create the session on writes, we'll also need to consider what options are available to tinker with things, e.g. to set the applicable domain for the cookie.

cburgdorf:
> so perhaps we should only create the session on writes

:+1: 

>e.g. to set the applicable domain for the cookie

Yep, that's a good point. Actually I will need to be able to set the domain for the use case that I currently have.

cburgdorf:
Hey @Ryman, I have the server that I'm writing as far that it basically does what I want it to do but with a `sessionid` right in the URL (actually it's even a valid GitHub OAuth token - tell nobody ;)). So, now I wanted to migrate that to use this branch and rely on a session cookie instead. But whoa it's giving me headaches. I wonder if you may be willing to give me a helping hand? It's currently in a private repository but I could give you access to it.

Ryman:
Closing in favor of #272

PR #254, opened 2015-08-10
Remove 1.0 build target from travis

Ryman:
hyper is tracking stable, so to keep things simple as rust develops, we should focus on the stable release channel as our moving target.

@cburgdorf @SimonPersson any hesitation?


cburgdorf:
Fine by me :+1:

SimonPersson:
Fine by me too!

PR #257, opened 2015-08-12
feat(response): allow templates without data

cburgdorf:
This improves ergonomics for template handling for when you don't want to pass data to the template. It's implementation relies on the already available `render` method (which I renamed to `render_with_data`) and just passes `&()`.

/cc @Ryman @SimonPersson

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -184,6 +184,23 @@ impl<'a> Response<'a, Fresh> {
         if !headers.has::<H>() { headers.set(f()) }
     }
 
+
+    /// Renders the given template
+    ///
+    /// # Examples
+    /// ```{rust}
+    /// use std::collections::HashMap;
+    /// use nickel::{Request, Response, MiddlewareResult, Halt};
+    ///
+    /// fn handler<'a>(_: &mut Request, mut res: Response<'a>) -> MiddlewareResult<'a> {
+    ///     res.render("examples/assets/template.tpl")
+    /// }
+    /// ```
+    pub fn render_without_data<T, P>(self, path: P) -> MiddlewareResult<'a>
+            where P: AsRef<str> + Into<String> {
+        self.render(path, &())
+    }
+
     /// Renders the given template bound with the given data.
     ///
     /// # Examples


SimonPersson:
But if you don't want any data, why would you use templates at all? Why not
send a plain HTML file? Just trying to understand the use case.

cburgdorf:
Ah, you mean with the `send_file` API. Actually I didn't consider that but you are right.
However, that API is a bit unergonomic to use currently because it doesn't work with `& ' static str`. But that should be easy to fix I guess.

So currently, this works:

```rust
server.get("/login", middleware! { |_, res|
    return res.send_file(Path::new("public/login.tpl.html"));
});
```

But even then I wonder if we shouldn't go with `render` and `render_with_data` (even if `render` would basically just use `send_file` under the hood) as it feels more natural to me. Thoughts?

cburgdorf:
Let's discuss this further in https://github.com/nickel-org/nickel.rs/pull/258

Ryman:
@SimonPersson In mustache specifically, if you don't provide data then a block won't render as it isn't "truthy", but the rest of the template would render fine. So you could imagine if a logged in user is provided then show one thing, otherwise another thing, data would only need to be provided in one case. Passing such a template to `send_file` would have weird output (with the templating `{{..}}`). 

I'm not really in favor of making the common path more verbose, i.e. I think people are more inclined to have data if they are using a template and I think it should be explicit in some way if no data is being rendered.

cburgdorf:
Ah, interesting! How would you feel about the change if we name it `render` and `render_without_data` then?

cburgdorf:
I went ahead and updated the PR with the proposed naming.

Ryman:
Well, honestly I think we should be aiming to remove template rendering into an `extension` so that other people can see how they could wrap their own templating language if they wanted to. 

I understand the want to render with `&()`, but I'm unsure how prevalent it is, in your specific case, I'm guessing you were just using a template as straight html?

@cburgdorf Test is failing on the doctest (and will need to update the commit message to the inverse)

cburgdorf:
>Well, honestly I think we should be aiming to remove template rendering into an extension so that other people can see how they could wrap their own templating language if they wanted to.

Fully agree on that. It's just something that it's not high on my priority list for now.

>I understand the want to render with &(), but I'm unsure how prevalent it is, in your specific case, I'm guessing you were just using a template as straight HTML?

Yes, to be honest, I just picked `&()` so that I can rely on the existing code and build the dataless API on top. For my specific case, I can live with the `send_file` API for now as I just needed to render a html file (without mustache).

>@cburgdorf Test is failing on the doctest (and will need to update the commit message to the inverse)

I'm fine with closing this for now. We can revisit when we approach to move the functionality into an extension.

PR #258, opened 2015-08-12
feat(render): improve ergonomics of send_file

cburgdorf:
Relates to #257

@Ryman @SimonPersson

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -114,7 +114,8 @@ impl<'a> Response<'a, Fresh> {
     ///     res.send_file(favicon)
     /// }
     /// ```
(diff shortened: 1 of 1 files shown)


cburgdorf:
As mentioned in #257 I'm tempted to rename `render` to `render_with_data` and create `render` as an alias to this improved `send_file` method. Thoughts?

Ryman:
I'm pretty sure this bounds is strictly more general, so I've merged.

I don't think `render` with no data should pass to `send_file`, I highlighted a reason on #257

cburgdorf:
yay, it's been a while since I last contributed code to the project. Feels good :)

Issue #259, opened 2015-08-12
Create mounting middleware


cburgdorf:
Currently our `StaticFileHandler` middleware can't play out it's full strength because every path that one registers will become relative available to root (`/`). Which isn't alway what we want.

We need some sort of mounting middleware (e.g. https://github.com/iron/mount) that allows to bind the contents of a path to any subpath. E.g. you may want to bind `/css` to `my/area/css` which isn't possible right now.

SimonPersson:
I think you can stick the current static files middleware in the router and
have it on whatever path you want, no? Perhaps we could document that just
:)

cburgdorf:
You mean like [this](https://github.com/cburgdorf/nickel.rs/commit/d487ca017a957240ffe8530b7f499b00c81a6ee1)? This doesn't work :-/

cburgdorf:
ping @SimonPersson ;)

Would love to this sorted (either by documenting, fixing or building something) for an 1.0 release.

SimonPersson:
Yeah, I didn't think my previous comment through really, hehe :)

I'll give this a shot sometime next week if I'm not beaten to it. I just moved across the continent so I'm a bit busy the next couple of days.

Ryman:
@cburgdorf Can we close this now?

cburgdorf:
I haven't had the chance to integrate that in classroom yet but I assume it works and will reopen new issues that I may come across when I start with the integration.

PR #267, opened 2015-08-23
Add Mount middleware

SimonPersson:
Attempt at #259. 

This middleware takes a mount point and a `Middleware`. When invoked, the mount point is compared with the incoming request, and if they match the underlying middleware is invoked but with the request path rewritten to have the mount point as root.

Ping @cburgdorf


src/mount.rs
```diff
@@ -0,0 +1,84 @@
+use request::Request;
+use response::Response;
+use middleware::{Continue, Middleware, MiddlewareResult};
+
+use hyper::uri::RequestUri::AbsolutePath;
+
+pub struct Mount {
```
Ryman: Should be able to get static dispatch with the following:

```rust
pub struct Mount<M: Middleware + Send + Sync + 'static> {
    mount_point: String,
    middleware: M
}
```

(It may not be required to specify `+Send+Sync+'static` as they are implied by `Middleware` but I forget if that propagates)

src/mount.rs
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    ///
+    /// # Examples
+    /// ```{rust}
+    /// use nickel::{Nickel, StaticFilesHandler, Mount};
+    /// let mut server = Nickel::new();
+    ///
+    /// server.utilize(
+    ///     Mount::new("/static_files/",
+    ///                StaticFilesHandler::new("/path/to/serve/")
```
Ryman: May be worth introducing a trait `Mountable` and implement it for `Nickel` so we can have `server.mount(foo, bar)`.

src/mount.rs
```diff
@@ -0,0 +1,84 @@
...
+            mount_point: rewrite_mount_point(mount_point),
+            mounted_handler: Box::new(mounted_handler)
+        }
+    }
+}
+
+impl Middleware for Mount {
+    fn invoke<'mw, 'conn>(&'mw self, req: &mut Request<'mw, 'conn>, res: Response<'mw>)
+        -> MiddlewareResult<'mw> {
+        if let AbsolutePath(path) = req.origin.uri.clone() {
```
Ryman: Would be nice to only do the clone here if the inner condition is true

src/mount.rs
```diff
@@ -0,0 +1,84 @@
...
+            if path.len() >= self.mount_point.len()
```
Ryman: I think this can be `path.starts_with(self.mount_point)`

src/mount.rs
Ryman: Using `mem::replace` could be even nicer if possible :dart:

src/mount.rs
```diff
@@ -0,0 +1,84 @@
...
+                return ret_val;
+            }
+        }
+        Ok(Continue(res))
+    }
+}
+
+#[test]
+fn mount_point_rewrites() {
+    assert!(rewrite_mount_point("/") == "/");
```
Ryman: `assert_eq!` will have better messages on failure here

examples/mount.rs
```diff
@@ -0,0 +1,22 @@
...
+
+fn main() {
+    let mut server = Nickel::new();
+
+    server.utilize(Mount::new("/a/b/c", middleware! {
+        "Hello world."
+    }));
+
+    server.utilize(Mount::new("/1/2/3", middleware! {
+        "H3110 w0r1d."
```
Ryman: I'm not sure these first two usages show anything useful to readers, perhaps printing `req.path_without_query()` would make things clearer?

examples/mount.rs
```diff
@@ -0,0 +1,22 @@
...
+    }));
+
+    server.utilize(Mount::new("/static/files/", StaticFilesHandler::new("examples/assets/")));
+    server.utilize(Mount::new("/static/files/a", middleware! {
+        "No static file called a!"
```
Ryman: Maybe a comment saying this is fall-through behavior, either way this is more useful than the top two usages :+1: (to me at least)

Ryman:
Looks like it could do the job :+1: 

I did take a look at iron's version mentioned in #259 and after looking at the code I'm left wondering a bit if we could abstract the core logic and both benefit from improvements, something to consider for the future I guess.

As for the api, it does contrast a bit with iron's version where they can mount multiple things with the same mount instance, which I guess could potentially have efficiency benefits if mounting multiple things (while being less efficient for single mounts). I'm happy enough with single mounts for now, but I'm wondering if it's a substantial usecase.

Finally, I'm wondering if this is worth isolating in a separate crate? Or do we just hold off until we have infrastructure in place to make that easy (whenever that is)? Again, I'm probably happy enough either way currently.

SimonPersson:
@Ryman, I've been through all your comments and adressed everything except the unnecessary copy when the matching branch is not taken. I can't seem to find an elegant way to solve that. If you can, please tell :) My problem is that all solutions I can think of introduces an extra `if`, and I'd like to not introduce an extra branch where it is not really needed.

Ryman:
> My problem is that all solutions I can think of introduces an extra if, and I'd like to not introduce an extra branch where it is not really needed.

What's the concern with an extra branch?

examples/mount.rs
```diff
@@ -0,0 +1,18 @@
...
+use nickel::{Nickel, Mountable, StaticFilesHandler};
+
+fn main() {
+    let mut server = Nickel::new();
+
+    /*
+     * Fall-through behaviour, if StaticFilesHandler does not find a matching file,
+     * the request uri must be reset so that it can be matched against other middleware.
+     */
+    server.mount("/static/files/", StaticFilesHandler::new("examples/assets/"));
```
Ryman: Looking at this again, the rewrite rules would make this `"/static/files"`? In that case, does this mean that the mount will activate for paths such as `"/static/files_and_stuff/bar"` and pass the uri as `"_and_stuff/bar"`?

I'm also curious about the behavior of `"/"`, as the uri passed to the mounted middleware would be `""`  if the base route is caught (i.e. the client sends `"/"`)? I would have expected `"/"` to be passed?

I'm not really sure what motivates the rewrite rules, can you add a comment to `rewrite_mount_point` with expected behavior?

SimonPersson:
>>
My problem is that all solutions I can think of introduces an extra if, and I'd like to not introduce an extra branch where it is not really needed.
>
What's the concern with an extra branch?

No big concerns, just that it looks a bit ugly. If there's a way around it, I'd like to avoid it. Otherwise, no big deal.

>
Looking at this again, the rewrite rules would make this "/static/files"? In that case, does this mean that the mount will activate for paths such as "/static/files_and_stuff/bar" and pass the uri as "_and_stuff/bar"?
>
Im also curious about the behavior of "/", as the uri passed to the mounted middleware would be "" if the base route is caught (i.e. the client sends "/")? I would have expected "/" to be passed?

Good catch, that's probably right! It's not really related to the rewrite rules though, even without those, this same issue would arise. If the mount path is "/static/files", then we also match "/static/files_and_stuff". If the mount path is "/static/files/", then a request to "/static/files/abc" ends up with the path "abc" where I'd expect it to be "/abc". I think this will work if we make sure that mount paths always end with a slash, and then prepend a slash to the path before calling the underlying middleware. That should solve both those failing cases.

>
I'm not really sure what motivates the rewrite rules, can you add a comment to rewrite_mount_point with expected behavior?

I'm not sure about them either. We can also just document which format mount points must be written in and hope that users get it right. Or, we could check for leading and trailing slash and return an error if the user got it wrong.

Ryman:
> Or, we could check for leading and trailing slash and return an error if the user got it wrong.

Assuming no-one is trying to mount things on an already running server I think this is a good option which allows people to fail *quickly* at runtime so they can get their stuff working with least surprise. Does make the usage kind of ugly though if it requires an unwrap, would this be a good place to just panic on the basis that it's most likely not what the user intended?

cc @cburgdorf

src/mount.rs
```diff
@@ -0,0 +1,95 @@
...
+            mount_point: rewrite_mount_point(mount_point),
+            middleware: middleware
+        }
+    }
+}
+
+impl<M: Middleware> Middleware for Mount<M> {
+    fn invoke<'mw, 'conn>(&'mw self, req: &mut Request<'mw, 'conn>, res: Response<'mw>)
+        -> MiddlewareResult<'mw> {
+        if let AbsolutePath(path) = req.origin.uri.clone() {
```
Ryman: *untested but might work and isn't totally ugly?*

```rust
let subpath = match req.origin.uri {
    AbsolutePath(ref path) if path.starts_with(self.mount_point) => {
        AbsolutePath(path[self.mount_point.len()..].to_owned())
    },
    _ => return Ok(Continue(res))
};

let original = mem::replace(&mut req.origin.uri, subpath);
let result = self.middleware.invoke(req, res);
req.origin.uri = original;
result
```

SimonPersson: Of course! I never thought to return early in the non mathing case! This will probably work and looks nicer than what I could think of :)

cburgdorf:
I've no idea why github doesn't let me add my comment where it belongs so I'll add it here.

>would this be a good place to just panic on the basis that it's most likely not what the user intended?

I'm absolutely in favor of the `panic!` here. It's probably the best way to notify the developer about a misconfiguration.

GitCop:

 * Commit: a023aa12520a076d1cfcafcb253d58933cd4631b

 * Commit: dc8043c022977c4386670f7f85c3a3096b06eb6e

 * Commit: 9169c1d0cdb30a66d896137fe5bb21d199b56be1

---

SimonPersson:
I've fixed the broken logic discovered by @Ryman, removed the rewrite rules and replaced them with a `panic!`, and fixed the unnecessary copy when the mount point does not match the request.

cburgdorf:
Yay! We finally have a bot! Love it :)

GitCop:

 * Commit: a023aa12520a076d1cfcafcb253d58933cd4631b

 * Commit: dc8043c022977c4386670f7f85c3a3096b06eb6e

 * Commit: 9169c1d0cdb30a66d896137fe5bb21d199b56be1

 * Commit: 3e7bda58147ef241e4d26b106d7e242d332105b1

 * Commit: 807566e99191d309cf63cac2c7b1969d767a6bb4

---

GitCop:

 * Commit: a023aa12520a076d1cfcafcb253d58933cd4631b

 * Commit: dc8043c022977c4386670f7f85c3a3096b06eb6e

 * Commit: 9169c1d0cdb30a66d896137fe5bb21d199b56be1

 * Commit: 3e7bda58147ef241e4d26b106d7e242d332105b1

 * Commit: 807566e99191d309cf63cac2c7b1969d767a6bb4

 * Commit: 41934476e99ec2e27964f3a9d60a1c8b963a709c

---

Ryman:
@SimonPersson If you reword commits or squash I think we're good to go with this for now? Or perhaps we should test @cburgdorf's project with commits from this branch?

SimonPersson:
I squashed the commits, let's see if the tests pass and you can feel free to merge it!

Ryman:
@homu r+

Ryman:
Err, it's definitely not commit 00000000

@homu r-

Ryman:
@homu r-

Ryman:
@homu r+

Ryman:
Sigh :D

Ryman:
@homu r+

Ryman:
Thanks @barosl! :surfer:

cburgdorf:
That's awesome! I'll test that with my classroom project later this week.

PR #269, opened 2015-08-24
feat(response): add next_middleware method for convenience

Ryman:
This seems more intuitive, less imports, less types to know about, does what it says.

### DIFF
--- a/examples/macro_example.rs
+++ b/examples/macro_example.rs
@@ -21,7 +21,7 @@ struct Person {
 //this is an example middleware function that just logs each request
 fn logger<'a, D>(request: &mut Request<D>, response: Response<'a, D>) -> MiddlewareResult<'a, D> {
     println!("logging request: {:?}", request.origin.uri);
-    Ok(Continue(response))
+    response.next_middleware()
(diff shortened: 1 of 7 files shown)


cburgdorf:
For ergonomics we could add the same for this, right?

```rust
let stream = try!(res.start());
Ok(Halt(stream))
```

Ryman:
Currently `res.send("")`?

cburgdorf:
True, but it's kind of unintuitive ;) I think having `res.halt_middleware()` feels more aligned with `res.next_middleware()`

Ryman:
I actually feel like we should change `res.end()` as it's not really needed anymore in it's current form (hyper now handles dropping the response properly)

Ryman:
@cburgdorf @SimonPersson Rebased, anyone want to r+? I think the `halt` discussion can go in a new thread?

SimonPersson:
Sure! Fix that one empty line and you can feel free to r+ yourself :+1:

Ryman:
@homu r=SimonPersson

PR #270, opened 2015-08-25
feat(router): add chainable routes

### LINKED ISSUE #262: Chainable routes
Is it possible to do something like this:

```rust
server.route("/api/v0/foo")
  .get(middleware!{ ... })
  .post(middleware!{ ... })
  .put(middleware!{ ... });
```

stephenmuss:
Closes #262

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,7 @@ path = "examples/moved_ownership.rs"
 
 name = "example_route_data"
 path = "examples/example_route_data.rs"
+
+[[example]]
+name = "chaining"
+path = "examples/chaining.rs"
--- a/examples/chaining.rs
+++ b/examples/chaining.rs
@@ -1,0 +1,18 @@
+#[macro_use] extern crate nickel;
+extern crate hyper;
+
+use hyper::method::Method;
+use nickel::{Nickel, HttpRouter};
+
+fn main() {
+    let mut server = Nickel::new();
+    server
+        .add_route(Method::Get, "/", middleware!("Hello World"))
+        .get("/get", middleware!("get"))
+        .post("/post", middleware!("post"))
+        .put("/put", middleware!("put"))
+        .patch("/patch", middleware!("patch"))
+        .delete("/delete", middleware!("delete"));
+
+    server.listen("127.0.0.1:6767");
+}
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -15,10 +15,11 @@ pub struct Nickel{
 }
 
 impl HttpRouter for Nickel {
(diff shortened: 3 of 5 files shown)


cburgdorf:
Thanks! Looks good to me. @Ryman @SimonPersson?

Ryman:
I guess the reason for not adding this to `add_route(..)` is to not break implementors? 

I think the current changeset is technically breaking for downstream too (although I don't think there is any downstream code overriding these methods) so I'm wondering if consistency is good to have here?

Worth noting that this slightly different from what #262 asks for (I also interpreted it as you have implemented until I rechecked now), where they want to set multiple verbs for a single route. I think this change is worth adopting, but I'm unsure if we should close #262 as it may still be useful to have such a thing?

E.g.
```rust
// this pr
server.get("/a/b/c", get_handler)
      .post("/a/b/c", post_handler)
      .delete("/a/b/c", delete_handler)

// vs 262
server.route("/a/b/c")
      .get(get_handler)
      .post(post_handler)
      .delete(delete_handler)
```

cburgdorf:
I think that #262 is worth implementing. However, isn't it possible to actually have both. Because #262 is operating on the result of a yet to be defined `route` method so it could have a different return type which should probably let us achieve both chaining methods independently, no?

To be honest, I consider #262 a little more useful than the chaining that is proposed with this PR. But I would be open to have both.

Regarding the breaking change. I'm not so much afraid of breaking changes at the moment as we are still in our `0.x` days ;)

Ryman:
@cburgdorf Yeah! Sorry, it was a quick response so I wasn't super clear but I was suggesting to take this now, do #262 later and have both. Also I'm in favor of the break, but would prefer it to be consistent with `add_route`! :)

cburgdorf:
Ah, I overlooked that `add_route` part :) I agree with that and also about the strategy to take this now and tackle #262 later. @stephenmuss Can you update the PR?

stephenmuss:
Thanks for the feedback. I'll update the PR.

stephenmuss:
@cburgdorf @Ryman I've updated the PR as per the comments.

As noted, the change to `add_route` will break downstream.

I did overlook the multiple verbs per route as discussed in #262. I'll see if I can have a crack at that next.

Ryman:
@stephenmuss Example looks good, would you mind adding "BREAKING CHANGE: HttpRouter methods now need to return self" to the commit message?

> I'll see if I can have a crack at that next.

Sounds great! Feel free to message if you run into any trouble :)

Other than that, LGTM.

stephenmuss:
@Ryman I've updated the commit message. Cheers.

Ryman:
@homu r+

Issue #161, opened 2015-03-05
Configuration options - production / development / bind port and so on


ninjabear:
It would be nice to have the ability to switch configurations in the same way that ruby/rails does.

Basically switching the mode to ```development``` switches off caching and puts out more detailed error messages (detailed 500s for example). 

Switching off caching means that you can work on front-end stuff like templates without recompiling and rerunning your application (the most extreme example being something like [the shotgun gem](https://rubygems.org/gems/shotgun)). And in turn production modes switch on all the caching and clever stuff for production use.

The bind port/ip isn't really a big issue but again, sometimes your production port may differ to whatever one you're messing around with while developing. 

This is a fairly low priority item imo.

SimonPersson:
It'd be nice to store all static files in the binary when building for production too. Simplifies deployment a bit if you don't have to keep the folder structure and so on.

ninjabear:
@SimonPersson agree. Java has a resources/ folder, something like that would be nice (though I appreciate rust doesn't produce a zip)

SimonPersson:
I mean, you can already do this to some extent. Provided you don't have too many static files to serve, this works:
```
router.get("/index.htm", middleware! {
    include_str!("static/index.htm")
});
```

I bet you could also write a macro that creates a route like that for every file in a directory. We just need to fix up the templates.

PR #277, opened 2015-09-05
feat(nickel): add options to Nickel, allowing the user to disable the messages output on listen

ixjf:
This is in relation to the pull request #124 - it adds two methods to Nickel - a static default_options, and set_options. The only option available is output_on_listen, which when disabled, won't display "Listening on x.x.x.x:xxxx/Press Ctrl+C to shutdown the server" (it is enabled by default, which is the current behavior). This could be extended later on for whatever other options.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@ extern crate modifier;
 #[macro_use] extern crate lazy_static;
 
 pub use nickel::Nickel;
+pub use nickel::Options;
 pub use request::Request;
 pub use response::Response;
 pub use middleware::{Action, Continue, Halt, Middleware, ErrorHandler, MiddlewareResult};
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -8,10 +8,15 @@ use hyper::status::StatusCode;
 //pre defined middleware
 use default_error_handler::DefaultErrorHandler;
 
+pub struct Options{
+    pub output_on_listen: bool,
+}
+
 /// Nickel is the application object. It's the surface that
 /// holds all public APIs.
 pub struct Nickel{
     middleware_stack: MiddlewareStack,
+    options: Options,
 }
 
 impl HttpRouter for Nickel {
@@ -33,7 +38,39 @@ impl Nickel {
         // they don't like the default behaviour.
         middleware_stack.add_error_handler(DefaultErrorHandler);
 
-        Nickel { middleware_stack: middleware_stack }
+        Nickel { middleware_stack: middleware_stack, options: Nickel::default_options() }
+    }
+    
+    /// Gets the default options for the framework
+    ///
+    /// # Examples
+    /// ```{rust}
+    /// # #[macro_use] extern crate nickel;
+    /// # fn main() {
+    /// use nickel::{Nickel, Options};
+    /// let mut server = Nickel::new();
+    /// let options = Options{output_on_listen: false, .. Nickel::default_options()};
+    /// server.set_options (options);
+    /// # }
+    /// ```
+    pub fn default_options() -> Options {
+        Options {output_on_listen: true}
+    }
+    
+    /// Sets the options for this instance
+    ///
+    /// # Examples
+    /// ```{rust}
+    /// # #[macro_use] extern crate nickel;
+    /// # fn main() {
+    /// use nickel::{Nickel, Options};
+    /// let mut server = Nickel::new();
+    /// let options = Options{output_on_listen: false, .. Nickel::default_options()};
+    /// server.set_options (options);
+    /// # }
+    /// ```
+    pub fn set_options(&mut self, options: Options) {
+        self.options = options;
     }
 
     /// Registers a middleware handler which will be invoked among other middleware
@@ -143,8 +180,10 @@ impl Nickel {
         let server = Server::new(self.middleware_stack);
         let listener = server.serve(addr).unwrap();
 
-        println!("Listening on http://{}", listener.socket);
-        println!("Ctrl-C to shutdown server");
+        if self.options.output_on_listen {
+            println!("Listening on http://{}", listener.socket);
+            println!("Ctrl-C to shutdown server");
+        }
     }
 }
 


src/nickel.rs
```diff
@@ -33,7 +38,39 @@ impl Nickel {
...
+    /// ```{rust}
+    /// # #[macro_use] extern crate nickel;
+    /// # fn main() {
+    /// use nickel::{Nickel, Options};
+    /// let mut server = Nickel::new();
+    /// let options = Options{output_on_listen: false, .. Nickel::default_options()};
+    /// server.set_options (options);
+    /// # }
+    /// ```
+    pub fn default_options() -> Options {
```
Ryman: Is there any benefit to this over implementing `Default` and having `Options::default()` instead?

src/lib.rs
```diff
@@ -15,6 +15,7 @@ extern crate modifier;
 #[macro_use] extern crate lazy_static;
 
 pub use nickel::Nickel;
+pub use nickel::Options;
```
Ryman: nit: `pub use nickel::{Nickel, Options};`

src/nickel.rs
```diff
@@ -33,7 +38,39 @@ impl Nickel {
...
+    /// ```{rust}
+    /// # #[macro_use] extern crate nickel;
+    /// # fn main() {
+    /// use nickel::{Nickel, Options};
+    /// let mut server = Nickel::new();
+    /// let options = Options{output_on_listen: false, .. Nickel::default_options()};
+    /// server.set_options (options);
+    /// # }
+    /// ```
+    pub fn set_options(&mut self, options: Options) {
```
Ryman: I think this should be `options_mut` as that would allow `get` or `set` behaviour?

src/nickel.rs
```diff
@@ -33,7 +38,39 @@ impl Nickel {
...
+    /// ```
+    pub fn default_options() -> Options {
+        Options {output_on_listen: true}
+    }
+    
+    /// Sets the options for this instance
+    ///
+    /// # Examples
+    /// ```{rust}
+    /// # #[macro_use] extern crate nickel;
```
Ryman: fwiw: These doctests can just be:
```rust
use nickel::{Nickel, Options};
let options = ..
server...
```

By default if you don't specify `fn main` or `extern crate` then everything gets wrapped in `extern crate $crate; fn main() { $doctest }`.

Ryman:
@ixjf What do you think about making the `Options` have a builder style api? 

```rust
let server = ..;
server.options_mut()
      .print_on_listen(false);
```

src/nickel.rs
ixjf: I guess not. I don't have a lot of knowledge in Rust yet, I was not aware of the Default trait.

src/lib.rs
ixjf: Indeed, thanks. I do know about the syntax, but didn't think about it there.

src/nickel.rs
ixjf: If this is ever to be used for other options, then yes, that would be best. For output_on_listen only, that wouldn't be very useful. I'll make the required changes.

src/nickel.rs
ixjf: I see, thanks. Again, I didn't know about this, I followed the comments for the function below that one.

ixjf:
It's definitely cleaner. I don't see why not.

ixjf:
@Ryman How would you go about implementing accessors, since you don't have function overloading in Rust?

Ryman:
I think for now though we don't need to expose `getters` and see if a need appears? Perhaps we should raise `print_on_listen(bool)` to be on `Nickel` instead and we can forget about `Options` until it has more purpose? (Perhaps in relation to #161)

We can just keep the flag as a private detail for now and see if anybody has a reason to read it other than `Nickel` itself (which can access the private data)? (I'm unsure about the best approach here, but it therefore it seems safer to expose less and do the simplest thing)

> since you don't have function overloading in Rust

Yeah, different functions would be the usual strategy. 

cc @cburgdorf

ixjf:
The only problem I see with that is that if you eventually want to add other options, you'd either have to stick them all in Nickel, have them in multiple places, or break compatibility.

Ryman:
You're right, it's a pity we don't have stability markers outside of `rustc`. I'm not too concerned with breaking _very minor_ things, it should be easy enough to migrate off a simple flag like this, especially if we haven't given enough api to let people to depend on reading the value. 

If we end up with something similar to the current code in future then `Nickel::print_on_listen(bool)` can just hopefully just wrap `self.set_options(..)`.

cburgdorf:
Sorry, I'm a hard time following (it's moment's like this when I recognize that I'm still a Rust noob). Where is the problem with 

```rust
server.options_mut()
      .print_on_listen(false);
```

With the `options_mut` approach, would we actually have `get/set` behavior available, no?

ixjf:
I don't have a lot of knowledge in Rust either, but I don't think that it would be possible to do cleanly as you would in a language like C++ where function overloading is supported. The problem with that approach is that you can't make print_on_listen both a setter and a getter, while allowing setter calls to be chained (or really, you simply can't make it both a setter and a getter - unless you make them a field instead of a function, but then it can't be chained).

Ryman:
@cburgdorf 
> With the options_mut approach, we would actually have get/set behavior available, no?

Only if exposing `Options` fields publicly would you get `get`, which would break code matching on it pretty easily if fields are added (so adding all the extra code doesn't seem worth it to me, as we don't know how the future will look here).

> Is it convention that Rust libraries use 0.x for non-dev releases, or are all libraries simply not stable enough?

Not sure what you mean by non-dev releases, but yes a lot of libs seem to be sticking with 0.x holding out for whatever rust feature/dependency they require to get the feature they want for 1.0! We've been trying to follow this although we've been quite slow at bumping actual new releases to crates.io (the git repo itself should be considered very unstable). We've been bumping the minor revision each time we've been breaking something substantial as with most of our dependencies, e.g. you can depend on 0.6.x to hopefully not break you if you depend on 0.6.0, but 0.7.0 can be a huge change (due to pre 1.0 experimentation) :)

cburgdorf:
Ok, I had to get my hand on some code to follow and now I actually see what's the problem. It's tricky if you wand to have chaining, too. I thought you could do something like this but it also doesn't work and is most likely nonsense that just seemed to be a good idea in my overly tired brain ;)

```rust
struct Options {
  some_bool: bool,
  some_string: String
}


struct Modifier<'options, T: 'options> {
  options: &'options Options,
  val: & 'options mut T
}

impl<'options,T> Modifier<'options, T> {
  fn set (&mut self, val: T) -> & 'options mut Options {
    self.val = val;
    &mut self.options
  }
}

impl Options {
  fn some_bool_mut (&mut self) -> Modifier<bool> {
    Modifier {
      options: self,
      val: self.some_bool
    }
  }

  fn some_string_mut (&mut self) -> Modifier<String> {
    Modifier {
      options: self,
      val: self.some_string
    }
  }

  fn default () -> Options {
    Options {
      some_bool: true,
      some_string: "Foo".to_owned()
    }
  }
}

fn main() {
  println!("Hello, world!");

  let options = Options::default();

  options
    .some_bool_mut()
    .set(false)
    .some_string_mut()
    .set("bar".to_owned());
}
```

Ryman:
@cburgdorf See https://github.com/reem/rust-modifier for a more full-blown implementation of what you've described (we're using that for some parts of `Response`, see [set](http://docs.nickel.rs/nickel/struct.Response.html#method.set)). I'm not sure if this single option warrants going all-in on adding that to `Nickel`, but perhaps I could be convinced?

ixjf:
I don't really like it that you have calls to different objects in the chain.

cburgdorf:
@Ryman haha, great! So it actually made at least a little bit of sense. I've stumbled over this in our nickel code before but never made the stretch to understand what it actually is :D

It's another thing to put on my list of things that I should learn about in Rust land.

Well, if we use it for `Response` already and it's not a super heavy and hairy thing to introduce to this part of the code, why not?

cburgdorf:
@ixjf that just seems to be a weakness in the crappy code that I made up. With the real `rust-modifier` thing used, it looks much cleaner: http://docs.nickel.rs/nickel/struct.Response.html#method.set

ixjf:
Ah, yes. That was actually what I was suggesting earlier.

Ryman:
> Well, if we use it for Response already and it's not a super heavy and hairy thing to introduce to this part of the code, why not?

Just seems less 'obvious' for users, but I guess this feature is probably only going to be used by people who know what they're doing. I don't have anything much against it, if you guys are happy to roll with this then that's good enough for me :+1:

ixjf:
Perhaps this should be left for when more options are to be added? It's getting too complicated for such a simple thing.

cburgdorf:
Yes, I agree. We can go for the simple solution now and reach for `rust-modifier` when we have more options.

ixjf:
So - does this stay as it is and this is dealt with later, or do we move forward and merge this (after making the changes we talked about)?

Ryman:
I'm happy enough for the simple solution (the method on nickel) to land for now, if you wanted to add a *this is an unstable method* to the doc comment then that would be pretty great too. Perhaps we should also name it `set_print_on_listen(bool)` as that seems more akin to common style.

Ryman:
Relevant to this discussion: https://github.com/carllerche/mio/commit/8b84f3307a94ddf9e5e2aca6a39663e00705f3af

ixjf:
This is the same that we discussed before, but it wouldn't allow getters. But to be honest, I'm not sure there's any option even worth getting by an application?

Ryman:
@ixjf Yeah, the link is very similar to how we discussed, which is good. It's just an example to see that this style is quite idiomatic (expert rust developer, prolific rust project).

> I'm not sure there's any option even worth getting by an application?

I think we should leave off getters until someone presents a need for it :) Then we can discuss the problem with more understanding of requirements.

Ryman:
@ixjf Any update on this?

euclio:
@ixjf Is this still in the works?

Ryman:
Closing as we've merged #310 which is based on this.

Thanks for the initial work and discussion on this feature!

PR #272, opened 2015-08-27
Another day, Another session.

Ryman:
After some back and forth and testing with @cburgdorf's project, this takes an alternative approach from #246 which should cause less dancing with borrowck while maintaining most of the functionality.

I also added some cargo `features` for the crate, which if you check `cookies_example` allows using non-encrypted cookies without any of the setup involving the datatype. It also provides a *decent* default implementation for the `KeyProvider` which will hopefully mean people aren't incentivized to use zerod-keys for convenience when they do want encryption/sessions.

I'm a little bit worried about the api for the cookies, as I did waste a bit of time when I was trying to `find` a cookie from within the `res.cookies()`  and never getting a result (as I should have been checking the inbound cookies on `req`). I had considered blocking the `find` functionality for the `Response`'s `CookieJar`, but I can imagine that it may be a requirement for some middleware to re-write an outgoing cookie for some reason. Perhaps `res.cookies_out()` and `req.cookies_in()`? It would really have been nice if `find` just fell through from the outgoing cookies to the incoming (new cookies override) but that might be unexpected by users.

PR #280, opened 2015-09-09
feat(server): shared data between requests and plugins for Response

Ryman:
I've ripped out the core from #272 and decided we should just push separate crates for now until the implementation for Session/Cookies has settled. This allows more breaking changes to the session/cookies implementations as we experiment without requiring a load of version bumps to nickel itself. It's more work for us to manage, but should be better for users :+1: 

I've also added a default type for the typeparam, so breakage in client crates should be minimal (until they start using the dataparam!) and only really affect people who've written library level stuff.

r? @cburgdorf @SimonPersson

### DIFF
--- a/examples/macro_example.rs
+++ b/examples/macro_example.rs
@@ -19,13 +19,13 @@ struct Person {
 }
 
 //this is an example middleware function that just logs each request
-fn logger<'a>(request: &mut Request, response: Response<'a>) -> MiddlewareResult<'a> {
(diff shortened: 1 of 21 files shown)


Ryman:
Pushed cookies to https://github.com/nickel-org/cookies

Will push session later after I've cherry-picked from old branch, have to head out for a bit first!

cburgdorf:
:+1: awesome. Can't wait for it to land :)

Ryman:
> Can't wait for it to land :)

== r+ ? :see_no_evil: (Let homu know if so!)

cburgdorf:
Let's do it :)

@homu r+

PR #281, opened 2015-09-11
feat(macros): allow hinting the server data type in middleware macro

Ryman:
This can sometimes be required when using some middleware which are
predicated on the datatype of the Server. An alternative would be to
litter the handler with extra type annotations (which is awkward without
type ascription), or to use explicit type-annotated functions as middleware.

Example usage:
```
middleware! { |res| <ServerData>
   // res is of type Response<ServerData>
}
```

### DIFF
--- a/src/macros/middleware.rs
+++ b/src/macros/middleware.rs
@@ -23,8 +23,29 @@
 /// server.listen("127.0.0.1:6767");
 /// # }
 /// ```
+///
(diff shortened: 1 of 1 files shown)


Ryman:
@cburgdorf This commit got forgotten in #280 :scream:

cburgdorf:
@homu r+

PR #287, opened 2015-10-13
doc(mustache): add example using helper functions

Ryman:
Add a basic example of how to use mustache's more complicated features. I haven't touched the api on `Response` as I'd prefer that we started to migrate the mustache specific stuff out of the core.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,12 @@ path = "examples/example_template.rs"
 
 [[example]]
 
+name = "template_helpers"
+path = "examples/template_helpers.rs"
+
+
+[[example]]
+
 name = "macro_example"
 path = "examples/macro_example.rs"
 
--- a/examples/assets/template.tpl
+++ b/examples/assets/template.tpl
@@ -2,4 +2,7 @@
     <h1>
         Hello {{ name }}!
     </h1>
+    {{#helper}}
+        Hello {{ name }} from a helper function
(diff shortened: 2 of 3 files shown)


GitCop:

 * Commit: dbe07b9bc7358020e138fc836aca47ac729f69f1
  * Invalid type. Valid types are **feat, fix, docs, style, refactor, perf, test, chore, revert**

---

SimonPersson:
I think you should cache the template manually if you don't want to use the api on `Response` (which I agree with).

Something like this:
```
let template = mustache::compile_path("examples/template.tpl").unwrap();
let mut server = Nickel::with_data(template);
```
followed by:
```
req.server_data().render_data(&mut stream, &data);
```
in the handler is nicer, no?

Ryman:
I mostly agree about the caching (but this was mainly just an example so people could do things rather than continue asking), but I think it can be much nicer and largely automatic, I should just get to implementing it. I've been wondering if I should make `nickel_templates` or just `nickel_mustache`, I think the simpler case for now is just the latter. Maybe it'll inspire someone else to implement `nickel_handlebars`!

The usage within the handler should probably still look like `res.render_data("template path", &data)` though. (The extension trait shouldn't need to be on the data itself?)

Ryman:
Closing in favor of implementing it out of core: https://crates.io/crates/nickel_mustache

PR #288, opened 2015-10-24
feat(router): Add chainable routes

mattbanister:
This doesn't work yet but I'm putting up the PR because I could use some help with it.


src/router/chainable_router.rs
```diff
@@ -0,0 +1,71 @@
...
+    path: String,
+    nickel: &'a mut Nickel<D>
+}
+
+pub trait ChainablePath<'a, D> {
+    fn route(&'static mut self, path: &str) -> ChainableRoute<'a, D>;
+}
+
+impl<'a, D: Sync + Send + 'static> ChainablePath<'a, D> for Nickel<D> {
+    fn route(&'static mut self, path: &str) -> ChainableRoute<'a, D> {
```
Ryman: These should work with `&'a mut self` rather than `&static ..`

mattbanister:
@Ryman Thanks very much for the help.  Seems to be working now.   Please take a look again when you have a chance and let me know if you have any feedback.  I'm still very much new to rust.  So, any feedback is appreciated.

One question I had was whether Rust can distinguish between trait methods with the same name but different arity.  When I try to import both my chainable router impls and the HttpRouter ones, I get a compilation error that the "get" method for example exists in both traits.  So I'm thinking it can't tell that they have different signatures/arities.  Maybe this is just a limitation of rust at the moment?

cburgdorf:
>So I'm thinking it can't tell that they have different signatures/arities

Yep, there's no such thing as method overloading in Rust and it's probably unlikely to be implemented.

However, you can always just shift syntax to specifically tell the compiler which method to invoke.

So, instead of writing this which would be forbidden:
```
use trait_with_get;
use other_trait_with_get;

some_thing.get();
some_thing.get(true); //compiler error
```

You could write:

```
use trait_with_get;
use other_trait_with_get;

trait_with_get.get(some_thing);
other_trait_with_get.get(some_thing, true);
```

I didn't test it out but I'm pretty sure it should work.

Ryman:
What @cburgdorf is referring to is [UFCS](https://github.com/rust-lang/rust/blob/b97b0e3929030693f5cf0498e5639580a8089fdf/src/doc/trpl/ufcs.md) but it should be `SomeTrait::method(instance_of_self, other_args,..)` rather than using `.` :)

You shouldn't really be running into the method name clash unless you've implemented both traits for a given type though? (has the code changed?) Inference should be able to see that the return from `route` has a `get` that is very different from the `HttpRouter::get`.

mattbanister:
@Ryman @cburgdorf 

> You shouldn't really be running into the method name clash unless you've implemented both traits for a given type though? (has the code changed?) Inference should be able to see that the return from route has a get that is very different from the HttpRouter::get.

Right, that's true.  I confused myself.  I had an intermediate implementation where I was just adding the last path added with `route` to the Nickel stuct (which is ugly anyway).  In that case, all of the methods were implemented on `Nickel` rather than `ChainableRoute`.  My question actually doesn't apply to this PR.  Thanks for the help guys!

mattbanister:
@Ryman @cburgdorf Is there anything I can do to improve this?  Any feedback is appreciated as I'm new with rust.

src/router/chainable_router.rs
```diff
@@ -0,0 +1,71 @@
+use hyper::method::Method;
+use middleware::Middleware;
+use nickel::Nickel;
+use router::http_router::{HttpRouter};
+
+pub struct ChainableRoute<'a, D: Sync + Send + 'static = ()> {
+    path: String,
```
Ryman: I think this can be `&'a str`, or you might need to introduce a new lifetime `'b` if the `route` chaining is to work (haven't tried)

src/router/chainable_router.rs
```diff
@@ -0,0 +1,71 @@
...
+impl<'a, D: Sync + Send + 'static> ChainablePath<'a, D> for Nickel<D> {
+    fn route(&'a mut self, path: &str) -> ChainableRoute<'a, D> {
+        ChainableRoute {
+            path: path.to_string(),
+            nickel: self
+        }
+    }
+}
+
+pub trait ChainableHandler<'a, D> {
```
Ryman: I think these methods could just be inherent methods, as it seems a bit much to require two trait imports for one feature?

examples/chainable_router.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+    let mut server = Nickel::new();
+    server
+        .route("/")
+        .get(middleware!("get"))
+        .post(middleware!("post"))
+        .put(middleware!("put"))
+        .patch(middleware!("patch"))
+        .delete(middleware!("delete"))
+        .route("/chain1")
```
Ryman: I'm unsure about chaining `route` here, it feels a bit at odds with rust's ownership model here, where there's a non-obvious change to the receiver in the middle of the chain.

I find it a lot less clear than splitting it out into:
```rust
server.route("/")
      .get(...)
      .post(..);

server.route("/chain1")
      .get(..)
      .post(..);
```

Perhaps it's worth not having the feature at first and then seeing if the ask appears in future?

src/router/chainable_router.rs
```diff
@@ -0,0 +1,71 @@
+use hyper::method::Method;
+use middleware::Middleware;
+use nickel::Nickel;
+use router::http_router::{HttpRouter};
+
+pub struct ChainableRoute<'a, D: Sync + Send + 'static = ()> {
+    path: String,
+    nickel: &'a mut Nickel<D>
```
Ryman: Wondering if this should be more generic so it can handle:

```rust
let mut x = Router::new();
x.route(..)
 .get(..)
 .post(..);
```

I think if people want to structure their app they probably want to deal with subrouters (which they can mount at certain endpoints) rather than passing around `Nickel` instances.

Ryman:
@mattbanister Sorry for the delay, I've left some feedback :) 

At a higher level though, do you think you might want to publish this as a crate? I think it builds nicely around the core so I think it would be a good example of how to make extensions to nickel if you wished to do so!

src/router/chainable_router.rs
Ryman: Perhaps `&mut R where R: HttpRouter`

examples/chainable_router.rs
mattbanister: I agree chaining route here is a little weird.  I'll remove that.  That was kind of an afterthought and I didn't really think it through very clearly.

src/router/chainable_router.rs
mattbanister: Okay thanks I'll give that a try.

src/router/chainable_router.rs
mattbanister: Yeah I'm trying to hack around my limitations in understanding lifetimes.  I'll try that and see if I can get it working with an &str.

src/router/chainable_router.rs
mattbanister: I agree that subrouters are desirable.  I read in one of the other issues that someone was looking to nest routes.  I need to sit down with this for a bit and give it some thought though.

mattbanister:
Hi @Ryman 

Thanks for the feedback.  I think that would be great to make it a separate crate.  I'll work on that and your other comments this week.  I really appreciate the help!

src/router/chainable_router.rs
Ryman: Could definitely be a future feature! Feel free to ask if you need anything

PR #289, opened 2015-10-30
fix(serverdata): extend lifetime of return value of {Request|Response}::server_data

Ryman:
This is much more flexible!

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -44,7 +44,7 @@ impl<'mw, 'server, D> Request<'mw, 'server, D> {
         }
     }
 
-    pub fn server_data(&self) -> &D {
+    pub fn server_data(&self) -> &'mw D {
         &self.data
     }
 }
--- a/src/response.rs
+++ b/src/response.rs
@@ -279,7 +279,7 @@ impl<'a, D> Response<'a, D, Fresh> {
         }
     }
 
-    pub fn server_data(&self) -> &D {
+    pub fn server_data(&self) -> &'a D {
         &self.data
     }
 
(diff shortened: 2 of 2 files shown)


Ryman:
@cburgdorf @SimonPersson  This also highlights an oversight from when we merged the serverdata branch, for `Response`, both `data` and `server_data` were implemented! (I plan to remove `data` on next minor version bump)

Ryman:
@homu r+

cburgdorf:
What's the effect of this in practice? (just trying to understand and learn)

Ryman:
Previously:
```rust 
let foo = response.server_data();
let bar = foo.bar(); // something derived from foo (a reference, not owned data)
return response.send(bar) // error: response is currently borrowed
```
Would fail as `server_data`'s lifetime was borrowing the `Response` (via lifetime elision). i.e. the signature was actually `fn server_data<'a>(&'a self) -> &'a Data`.

This change, links the lifetime to something which outlives the `Response`, so by the function signature (non-elided) `fn server_data<'a>(&'a self) -> &'mw Data`, it's not tying the lifetime of the call (`'a`) to `server_data`. 

This means we're able to continue using the response  and are free to mutate and move it, while `server_data` is guaranteed to live as long as the lifetime `'mw` is valid (which is a lot longer than what it was before, which was linked to that specific function call).

cburgdorf:
Gotcha! Thank you for the insights. Greetings from Thailand btw ;)
Am 02.11.2015 05:52 schrieb "Ryman" <notifications@github.com>:

PR #290, opened 2015-10-30
fix(send_file): don't overwrite the ContentType header if it exists

### LINKED ISSUE #285: Ability to disable mimetype-guessing for send_file
The file has no file extension, but I know the mimetype.

Some ideas:

* Use `set_header_fallback` inside `send_file` so I can set the header beforehand
* Implement `Responder` for file objects, change `send_file` to build the mimetype guessing on top of that.

Ryman:
Fixes #285.


Ryman:
@homu r+

PR #295, opened 2015-11-09
fix(static_files): deny suspicious paths

Ryman:
Should really have done this a long time ago >_<

### DIFF
--- a/src/static_files_handler.rs
+++ b/src/static_files_handler.rs
@@ -4,6 +4,7 @@ use std::fs;
 
 use hyper::method::Method::{Get, Head};
 
+use status::StatusCode;
 use request::Request;
 use response::Response;
 use middleware::{Middleware, MiddlewareResult};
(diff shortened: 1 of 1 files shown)


cburgdorf:
So, basically this prevents from serving directories outside of the server directory, right? Is this common sense for web servers? What if I want to serve a directory outside of the server directory? Is this a security issue per se?

Ryman:
> So, basically this prevents from serving directories outside of the server directory, right?

Not exactly, it only puts restrictions on the path which comes from the user (client). The server can still use any path they want for the `base_path`.

> Is this common sense for web servers? Is this a security issue per se?

Yes, I think we should probably start to formalize some sort of process for informing users? I'll leave you a pm on gitter.

Ryman:
@cburgdorf ping on this, I don't think you've seen the gitter message yet?

PR #297, opened 2015-11-18
perf(hyper): re-enable keep-alive by default

Ryman:
This was disabled in hyper 0.6.15 which gave a significant performance regression. 

In the same release though `socket_timeouts` related code was included which allows us to add timeouts on the socket which should prevent threads being stalled indefinitely. With the `timeouts` features enabled this should match 0.6.14 performance and not lead to exhausted servers (although it can still happen, but they will recover, unlike before).

When the `timeouts` feature is disabled (to compile on 1.3), the performance should match the previous, but the keep-alive issues that were previously present will still occur, so it's probably best for users of 1.3 to either upgrade or to set `keep_alive_timeout` to `None`. 

I'm unsure if we should continue to support 1.3 with such a limitation. (especially since 1.4 is now stable)

cc @ruizander @cburgdorf @SimonPersson

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,24 +5,25 @@ cache:
   directories:
       - target
 
-rust:
-  - stable
-  - nightly
-  - beta
-
 matrix:
   fast_finish: true
   allow_failures:
     - rust: nightly
+  include:
+      - rust: 1.3.0 # minimum supported rustc version
+        env: FEATURES="--no-default-features"
+      - rust: stable
+      - rust: beta
+      - rust: nightly
 
 before_script:
   - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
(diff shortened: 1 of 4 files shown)


cburgdorf:
Sounds good. @homu r+

>I'm unsure if we should continue to support 1.3 with such a limitation. (especially since 1.4 is now stable)

How would we not continue to support it? Can we prevent people from compiling with 1.3 or print a warning? Or do you mean just adding something to the readme.

Ryman:
> How would we not continue to support it? Can we prevent people from compiling with 1.3 or print a warning? Or do you mean just adding something to the readme.

If we don't compile, then we prevent them compiling :P In this case, it would be a non-optional dependency on hyper's `timeouts` feature, which uses things which only exist as stable in 1.4.

PR #304, opened 2015-11-30
doc(examples): add examples with custom error handling

flosse:


examples/error_handling.rs
```diff
@@ -0,0 +1,77 @@
...
+  Err(AppError::Custom("uups".to_string()))
+}
+
+fn will_work() -> AppResult<String> {
+  Ok("foo".to_string())
+}
+
+fn foo_handler<'a, D>(_: &mut Request<D>, mut res: Response<'a,D>) -> MiddlewareResult<'a, D> {
+
+  let x: AppResult<String> = (||Ok({
```
flosse: This is actually a hack that I really like to be simplified

Ryman:
Thanks for taking initiative on this!

I've been delaying a bit on this as I think it requires a minor breaking change to get working properly due to coherence. I'd like to get our current behavior a bit more fleshed out (with the help of https://github.com/nickel-org/nickel.rs/pull/299) so that I can know the real impact of the changes.

I've amended your branch here with an example of how it *could* work: https://github.com/nickel-org/nickel.rs/compare/master...Ryman:flosse-error-handling

flosse:
> I've amended your branch here with an example of how it could work

Thanks a lot!
Do you also have a working example that is compatible with a stable nickel version? Or any recommendations?
At the moment it seems to me that rusty people just don't care about errors and really like the `unwrap()` method ;-)

PR #309, opened 2016-01-10
feat(nickel): add method to set thread count

jolhoeft:
Add a method to override hyper's default thread count of five. Applications regularly calling XmlHttpRequest could easily tie up five threads if keep-alives are enabled.

### DIFF
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -16,6 +16,7 @@ pub struct Nickel<D: Sync + Send + 'static = ()> {
     middleware_stack: MiddlewareStack<D>,
     data: D,
     keep_alive_timeout: Option<Duration>,
+    thread_count: Option<usize>,
 }
 
 impl<D: Sync + Send + 'static> HttpRouter<D> for Nickel<D> {
@@ -48,7 +49,9 @@ impl<D: Sync + Send + 'static> Nickel<D> {
             middleware_stack: middleware_stack,
             data: data,
             // Default value from nginx
-            keep_alive_timeout: Some(Duration::from_secs(75))
+            keep_alive_timeout: Some(Duration::from_secs(75)),
+            // Use hyper's default
+            thread_count: None
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/nickel.rs
```diff
@@ -178,6 +181,20 @@ impl<D: Sync + Send + 'static> Nickel<D> {
...
+    /// Set the number of threads the server should use
+    ///
+    /// # Performance
+    ///
+    /// Increasing this can help with thread exhaustion. Setting this
+    /// to `None` will use the default value.
+    ///
+    /// # Default
+    ///
+    /// The default value is 5 threads
```
Ryman: I think the default is a calculation in hyper. Last I checked it was `num_cpus * 5 / 4` or something similar?

Ryman: nit: fullstop after

src/server.rs
```diff
@@ -49,6 +49,9 @@ impl<D: Sync + Send + 'static> Server<D> {
             server.keep_alive(timeout);
         }
 
-        server.handle(arc)
+        match thread_count {
```
Ryman: Perhaps we should be adding our own calculation here as the default instead of hypers, so we can keep our docs in sync? Or else we'll just have to provide a link to hyper-docs for the calculation?

jolhoeft: I think there is an advantage to being able to use whatever defaults hyper provides. I presume they have tuned it for the common use cases, and we are providing overrides for less common ones. I think we should just note in the docs that 'None' will use hyper's default. If they improve their default calculations, we can benefit without needing to do anything.

src/nickel.rs
jolhoeft: I didn't realize it was a calculation, was just going by the logs. I'll update the docs.

GitCop:

 * Commit: 9d5e37d919639cead4a6ac181e59dcda33fede56

---

Ryman:
Mind squashing the last commit?

cc @cburgdorf I'm happy to take this, but maybe landing something like #277 and then adding this option to that would be better overall?

jolhoeft:
Sorry about that. I'm still learning the github workflow. This is my first pull request. It should be squashed now.

jolhoeft:
Conflicts resolved

Ryman:
@jolhoeft Great! Rebase looks good to me, welcome to Github :+1: 

@cburgdorf ping about the question above :)

cburgdorf:
>I'm happy to take this, but maybe landing something like #277 and then adding this option to that would be better overall?

I agree, that would be better. But the PR seems to be stale. I'm happy to go either way.

SimonPersson:
@jolhoeft - I just merged a PR that adds options to Nickel, 7994ad4. Would you mind rewriting this to use those? We'd be happy to merge this then :)

jolhoeft:
Does the upcoming async support in hyper render this feature obsolete? If not I'll rework into the current code.

Ryman:
@jolhoeft I think even in the async world it could be desired to be able to run across multiple threads (that probably means multiple instances of the hyper event loop being run in parallel, but I'd guess they'll also add their own convenience helpers for that which we can wrap in future)

I'd be happy to try to rebase this PR if you can't find the time though. It might be worth getting some kind of regression test in place for this kind of thing.

Ryman:
This was rebased and merged in #335, thanks for the initial push here @jolhoeft!

PR #310, opened 2016-01-22
feat(nickel): add options to Nickel, redux

euclio:
The original PR (#277) appears to be dead. @ixjf

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@ extern crate modifier;
 #[macro_use] extern crate log;
 #[macro_use] extern crate lazy_static;
 
-pub use nickel::Nickel;
+pub use nickel::{Nickel, Options};
 pub use request::Request;
 pub use response::Response;
 pub use middleware::{Action, Continue, Halt, Middleware, ErrorHandler, MiddlewareResult};
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -10,12 +10,49 @@ use hyper::status::StatusCode;
 //pre defined middleware
 use default_error_handler::DefaultErrorHandler;
 
+/// Configuration options for the server.
+///
+/// This struct provides a builder-style API for constructing the desired options.
+///
+/// # Examples
+/// ```{rust}
+/// use nickel::{Nickel, Options};
+/// let mut server = Nickel::new();
+///
+/// // Don't print to stdout when starting the server.
+/// server.options = Options::default().output_on_listen(false);
+/// ```
+pub struct Options {
+    output_on_listen: bool,
+}
+
+impl Options {
+    /// Whether the server should print the local address it is listening on when starting.
+    ///
+    /// Defaults to `true`.
+    pub fn output_on_listen(mut self, output: bool) -> Self {
+        self.output_on_listen = output;
+        self
+    }
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Options {
+            output_on_listen: true,
+        }
+    }
+}
+
 /// Nickel is the application object. It's the surface that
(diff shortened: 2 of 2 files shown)


Ryman:
Sorry for the delayed reply! In the old thread, I was leaning towards just adding a method to `Nickel`. I think I'm still leaning that way for simplicity.. 

But if we were to adopt this options, then perhaps it's worth moving things like `keep_alive_timeout` into the `Options` and having the method `keep_alive_timeout` dealing with `options` internally (and add a comment that the old method on `Nickel` is depreciated I guess?)

Also, if we do adopt, I'd prefer a builder style interface for the Options struct else it can break for users who create one via `Options { foo: 3, bar: 4 }` whenever we add new options. Alternatively add at least one private field to the struct, which would prevent people from using that form, but then they would need to go through an awkward `let opts = Options::default(); opts.bar = foo;` dance so I still lean towards a builder.

@euclio Do you have an opinion as a user?
@cburgdorf, @SimonPersson This is blocking a final decision on #309, any suggestions?

Ryman:
:zap: *Friendly ping* :zap:

SimonPersson:
>But if we were to adopt this options, then perhaps it's worth moving things like keep_alive_timeout into the Options and having the method keep_alive_timeout dealing with options internally (and add a comment that the old method on Nickel is depreciated I guess?)

Sounds good.

>Also, if we do adopt, I'd prefer a builder style interface for the Options struct else it can break for users who create one via Options { foo: 3, bar: 4 } whenever we add new options. Alternatively add at least one private field to the struct, which would prevent people from using that form, but then they would need to go through an awkward let opts = Options::default(); opts.bar = foo; dance so I still lean towards a builder.

Couldn't `options` just be a public member of the `Nickel` struct? Options could be set with `server.options.output_on_listen = true;`. It's a simple change to this PR, won't break if we change the struct, doesn't require us to add a new method on the `Nickel` struct. A builder pattern works for me too though.

euclio:
No opinions here.

Ryman:
> Couldn't options just be a public member of the Nickel struct? Options could be set with server.options.output_on_listen = true;. It's a simple change to this PR, won't break if we change the struct, doesn't require us to add a new method on the Nickel struct. A builder pattern works for me too though.

As long as `Options` can't be constructed with a struct literal (by adding a private field) then I think this would be ok? 

```rust
// e.g. with a `move`-ing builder - ok
server.options = Options::default().foo(4).bar(5);
// struct literal - could break if new fields added
server.options = Options { foo: 4, bar: 5 };
```

SimonPersson:
Sounds good to me.

@euclio: Do you want to go ahead and do this? As far as I can tell, we want a builder for `Options`, a private field in `Options`, and making `options` public in `Nickel`.

euclio:
@SimonPersson sure, I'll take a look over the weekend.

euclio:
I'm not sure if GitHub notifies the maintainers when I add commits... Friendly ping.

SimonPersson:
@euclio: I'm not sure if we're notified by new commits, but I don't think you actually pushed a new commit? I only see the one from 26 days ago.

euclio:
@SimonPersson I amended the old commit to use a builder-style API as discussed.

SimonPersson:
@euclio: I see. Could you also add a private field to `Options`, so that it is not possible to do `Options{output_on_listen:false}`? Perhaps using the `PhantomData` type.

euclio:
`output_on_listen` is already private, so that won't compile outside of the Nickel module.

Ryman:
LGTM

SimonPersson:
@homu r+

Issue #264, opened 2015-08-16
feat(https): add API for HTTPS


arden:
there have any plans for https support?
because ios9 only support the https.

flosse:
There is a [closed pull request](https://github.com/nickel-org/nickel.rs/pull/200) that could be reopened and improved to support https.

cburgdorf:
I put this on the list for our 1.0 milestone.

arden:
now i'm modify the server.rs
impl Server {
    pub fn new(middleware_stack: MiddlewareStack) -> Server {
        Server {
            middleware_stack: middleware_stack,
            templates: RwLock::new(HashMap::new())
        }
    }

    pub fn serve<T: ToSocketAddrs>(self, addr: T) -> HttpResult<Listening> {
        let arc = ArcServer(Arc::new(self));
        let server = try!(HyperServer::http(addr));
        server.handle(arc)
    }

    pub fn serve_https<T: ToSocketAddrs>(self, addr: T, cert: &Path, key: &Path) -> HttpResult<Listening> {
        let arc = ArcServer(Arc::new(self));
        let ssl = Openssl::with_cert_and_key(cert, key).unwrap();
        let server = try!(HyperServer::https(addr, ssl));
        server.handle(arc)
    }

    pub fn serve_with_threads<T: ToSocketAddrs>(self, addr: T, n: usize) -> HttpResult<Listening> {
        let arc = ArcServer(Arc::new(self));
        let server = try!(HyperServer::http(addr));
        server.handle_threads(arc, n)
    }

    pub fn serve_https_with_threads<T: ToSocketAddrs>(self, addr: T, cert: &Path, key: &Path, n: usize) -> HttpResult<Listening>  {
        let arc = ArcServer(Arc::new(self));
        let ssl = Openssl::with_cert_and_key(cert, key).unwrap();
        let server = try!(HyperServer::https(addr, ssl));
        server.handle_threads(arc, n)
    }
}

and modify the nickel.rs
    pub fn listen<T: ToSocketAddrs>(mut self, addr: T) {
        self.middleware_stack.add_middleware(middleware! {
            (StatusCode::NotFound, "File Not Found")
        });

        let server = Server::new(self.middleware_stack);
        let listener = server.serve(addr).unwrap();

        println!("Listening on http://{}", listener.socket);
        println!("Ctrl-C to shutdown server");
    }

[...]

PR #316, opened 2016-03-09
Https support

Vaelden:
This PR adds support for HTTPS.

Nickel now has a method `Nickel::listen_https()` which takes a hyper SSL context as parameter.

The SSL support is currently not present by default. Should we make it a default feature?

See #264


GitCop:

 * Commit: 905bf2841465eb485482a95b27648c274f87bade
  * Invalid type. Valid types are **feat, fix, docs, style, refactor, perf, test, chore, revert, doc**

 * Commit: a97d3f8e7e142951dcb0d7835d5055cacdc955d0
  * Invalid type. Valid types are **feat, fix, docs, style, refactor, perf, test, chore, revert, doc**

 * Commit: d8ac1450ee0d46d0f3f753fe91153cbbeaa6d8de

---

examples/https.rs
```diff
@@ -0,0 +1,27 @@
+#[macro_use] extern crate nickel;
+extern crate hyper;
+
+#[cfg(feature = "ssl")]
+fn main() {
+    ssl::main();
```
Ryman: Any reason to not just inline things here?

src/nickel.rs
```diff
@@ -190,6 +190,58 @@ impl<D: Sync + Send + 'static> Nickel<D> {
...
+    /// use hyper::net::Openssl;
+    ///
+    /// let server = Nickel::new();
+    /// let ssl = Openssl::with_cert_and_key("foo.crt", "key.pem").unwrap();
+    /// server.listen_https("127.0.0.1:6767", ssl);
+    /// ```
+    impl <D: Sync + Send + 'static> Nickel<D> {
+        pub fn listen_https<T,S>(mut self, addr: T, ssl: S)
+        where T: ToSocketAddrs,
+              S: Ssl + Sync + Send + Clone + 'static {
```
Ryman: Is `Sync` required here? (corresponding api in [hyper](https://github.com/hyperium/hyper/blob/master/src/server/mod.rs#L217) requires only `Ssl + Clone + Send`)

examples/https.rs
Vaelden: Hmmm ... you just made me discover I could use `use` inside functions ... So no, no particular reason.

Ryman:
Thanks @Vaelden, this looks pretty good :+1: 

Can you try adding a basic test just to ensure things keep working in future? You can use [hello world](https://github.com/nickel-org/nickel.rs/blob/master/tests/examples/hello_world.rs)  as a reference and then add the module to [here](https://github.com/nickel-org/nickel.rs/blob/master/tests/example_tests.rs). The travis config will probably need changed, but I can take care of that later if it's causing trouble for you.

I'd like to reduce the duplication in the `listen_foo` methods, but the public api seems ok so I'm happy to leave that for now and clean it up in future since it's only internal. So, I don't think it's a blocker.

> Should we make it a default feature?

I'm open to this change, it was removed as a default before since we didn't actually have https, so perhaps if this is merged then being a default feature would make sense. I think the reason to make it optional was mainly for windows users and they can use `--no-default-features`  if they need. (although this should probably require a version bump? We require one anyway if we're to update hyper, which we should do soon)

Vaelden:
OK I will try to do all of that soon.

I also just spotted a problem with the doc tests, which fail when running `cargo test --features ssl`. I tried a few things but can't make it work properly ... rustdoc seems to do weird things with features (compiles the module, but with the feature disabled so it doesn't find `listen_https()`). However it does work properly (with a few modifications) when ssl is a default feature.

Ryman:
@Vaelden I played with this a bit to try to figure it out, the real mystery was `cargo test --features ssl` would fail, but `cargo test --features ssl ssl` (limiting to only run the doctest) would work. 

It's caused by a poor interaction with the example testing infrastructure, where they always recompile the examples (incase you only wanted to be running specific tests without rebuilding extra stuff). They try to build the examples with a set of features, but we don't / can't(?) currently pull the active features that the tests are built with, so I added a hack to grab them from an env-var that we use for travis. 

tl;dr; `export FEATURES="--features ssl"; cargo test --features ssl` should work.

Vaelden:
I see. It looks like the features being compiled are only exposed to the [build script](http://doc.crates.io/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) so maybe there would be a clever way to use that.

As for the test itself, I would also need to make sure that my example always gets compiled with `--features ssl`, which revealed a bit tricky. Should I assume that the test_example will get compiled with the *same features* as those present in `FEATURES`? (i.e. `FEATURES=--no-default-features cargo test --features ssl` would fail, and even `cargo test --features ssl` if ssl is not a default feature)

For the code of the example itself, I need to set the CA file to the same the server is using. To do that, I need to play with `openssl::ssl::SslContext`, but I wasn't sure if adding `openssl` as dev-dependency would be a good idea (problems on Windows?).

I found a workaround that is a bit ugly to avoid depending on openssl, should I use that instead? (see updated PR)

Ryman:
> As for the test itself, I would also need to make sure that my example always gets compiled with --features ssl, which revealed a bit tricky. Should I assume that the test_example will get compiled with the same features as those present in FEATURES? (i.e. FEATURES=--no-default-features cargo test --features ssl would fail, and even cargo test --features ssl if ssl is not a default feature)

This should always be true on travis (else it's a bug), it's just a pain to do locally currently, maybe a makefile is worth adding (not in this PR).

> I see. It looks like the features being compiled are only exposed to the build script so maybe there would be a clever way to use that.

Yeah, I just don't know of a good way to do that, especially if/when we end up ripping out the testing infrastructure so other crates can use it. I hacked together a local version of cargo last night that exposed them to tests as well as build scripts to see if it's possible (it is) but there might be good reason it's not done.

> For the code of the example itself, I need to set the CA file to the same the server is using. To do that, I need to play with openssl::ssl::SslContext, but I wasn't sure if adding openssl as dev-dependency would be a good idea (problems on Windows?).
> 
> I found a workaround that is a bit ugly to avoid depending on openssl, should I use that instead? (see updated PR)

If the ssl feature is active, then openssl is already going to get pulled in as a dependency due to hyper pulling it in? You could add it as an optional dependency (would like to have it as a dev-dependency but they can't be optional afaik). Fwiw, I don't think the workaround you have in the commit is that bad (apart from the `Arc::get_mut` dance that I don't fully understand), but I don't really know how the alternative looks.

Vaelden:
>If the ssl feature is active, then openssl is already going to get pulled in as a dependency due to hyper pulling it in? You could add it as an optional dependency (would like to have it as a dev-dependency but they can't be optional afaik).

You're right, the optional dev-dependency would have been the best way, but it is currently not possible. The optional dependency works, I just find it weird to depend on a crate just for one test ...

My workaround was just to get the `SslContext` generated by hyper, but as it is inside an Arc I needed to get a &mut reference to it, hence the `get_mut()` (that returns a &mut when there is exactly one reference to it). In my opinion, the proper way to do it would be:

```rust
fn create_context() -> Openssl {
    let mut ssl = SslContext::new(SslMethod::Sslv23).unwrap();
    ssl.set_CA_file("examples/assets/self_signed.crt").unwrap();
    Openssl { context: Arc::new(ssl) }
}
```

Anyway the current version works, so feel free to merge *as is*, or we can put openssl as optional dependency if you want to.

Ryman:
@SimonPersson This looks good to me (other than the duplication which we can clear up internally in future), any concerns?

SimonPersson:
Looks good to me too! Feel free to merge :)

Ryman:
@homu r+

Issue #326, opened 2016-04-05
Router doesn't match any routes after a POST, thinks request is an Extension instead of a POST


shepmaster:
Originally reported on [Stack Overflow](http://stackoverflow.com/q/36408373/155423).

## Reproduction steps

Create a new Cargo project

**Cargo.toml**

```toml
mustache = "0.7.0"
nickel = "0.8.0"
rustc-serialize = "0.3.19"
```

**main.rs**

```rust
#[macro_use]
extern crate nickel;
extern crate mustache;
extern crate rustc_serialize;

use std::collections::HashMap;
use nickel::{Nickel, MediaType, HttpRouter};
use nickel::status::StatusCode;

fn main() {
    let mut server = Nickel::new();
    let mut router = Nickel::router();

    router.get("/", middleware!(|request, mut response| {
        response.set(StatusCode::Ok);
        response.set(MediaType::Html);
        return response.send_file("assets/login.tpl");
    }));

    router.post("/login", middleware!(|request, mut response| {
        response.set(StatusCode::Ok);
        response.set(MediaType::Html);

        let mut data: HashMap<&str, &str> = HashMap::new();
        data.insert("error", "hello");
        return response.render("assets/login.tpl", &data);
    }));

    server.utilize(router);
    server.listen("127.0.0.1:6767");
}
```

**assets/login.tpl**

```
<html lang="en">
    <head>
        <meta charset="utf8"/>
    </head>
    <body>
        <h1>Login</h1>
        <form method="post" action="login">
            <label for="email">Email</label>
            <input type="email" name="email"/>
            <br/>
            <label for="password">Password</label>
            <input type="password" name="password"/>
            <br/>
            <button type="submit">Login</button><br/>
            <a href="/register">Register</a>
        </form>
        {{error}}
    </body>
</html>
```

1. Visit http://127.0.0.1:6767/login
2. Fill in the fields and submit
3. Fill in the fields a second time

What is expected: The form continues to be shown
What happens: The server returns an HTTP 404

## Initial investigation

I added some debug prints to **router.rs**:

```rust
.inspect(|item| println!("{:p}, {:?}, {:?}, {:?}", self, item.method, method, item.matcher.is_match(path)))
```

On the first POST, it iterates through the middleware with the following:

```
0x107e29160, Get, Post, false
0x107e29160, Post, Post, true
```

On the second POST, the incoming request's method is all screwy:

```
[...]

PR #327, opened 2016-04-10
fix(response): prevent keep-alive if the Request had a body

Ryman:
This puts a bandaid on #326. A more complete fix will require
some shuffling of the api exposed by Request so that we can
monitor for when the body has been fully read (or read up to
a defined limit).


Ryman:
Looks like Rust 1.6 is new minimum due to url bumping to new uuid version (which added no_std support). Will update the PR if this fix is desired.

PR #331, opened 2016-04-24
chore(travis): bump minimum rust version

Ryman:
Rust 1.7 is new minimum due to url bumping to new uuid version (which added no_std support) and cookie bumping to url 1.0 which depends on IpAddr which was destabilized in 1.6 and then restabilized in 1.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 
 rust:
-  - 1.5.0 # minimum supported rustc version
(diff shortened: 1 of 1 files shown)


Ryman:
Looks like cookie updated to url version 1.0 which depends on an even newer feature.

Ryman:
@cburgdorf Maybe with the trains running it's not worth tracking minimum version anymore?

Ryman:
@homu r+

PR #332, opened 2016-04-25
feat(urlencoded): add `application/x-www-form-urlencoded` parser

### LINKED ISSUE #329: Get `application/x-www-form-urlencoded` POST data as a HashMap
Hello! I would like to get `application/x-www-form-urlencoded` POST data as a HashMap. 
As I understand it's impossible at this moment, isn't it? Are there any plans to implement this?

Kilte:
Add FormBody trait which allows to extract URL encoded data from the request body.

Closes #329


examples/form_data/form_data.rs
```diff
@@ -14,15 +13,16 @@ fn main() {
...
+        let form_data = try_with!(res, req.post());
 
-        println!("{}", form_data);
+        println!("{:?}", form_data);
 
         let mut data = HashMap::new();
         data.insert("title", "Confirmation");
-        data.insert("formData", &form_data);
-
+        data.insert("firstname", form_data.get("firstname").unwrap_or("What's your first name?"));
```
Ryman: I guess there's no easy way to get this working with serialization so something like: 
```rust
let foo: SomeType = try_with!(req, form_data.parse())
``` 
would work? Probably something that can be left for future though.

src/lib.rs
```diff
@@ -23,7 +23,7 @@ pub use mount::{Mount, Mountable};
 pub use favicon_handler::FaviconHandler;
 pub use default_error_handler::DefaultErrorHandler;
 pub use json_body_parser::JsonBody;
-pub use query_string::{QueryString, Query};
```
Ryman: It's true that both of these are handling urlencoded things but I think it's slightly confusing to have to think about that when navigating the code. I'd prefer to capture the similarities within the urlencoded module but still leave `Query` and `FormBody` in their own modules.

src/urlencoded.rs
```diff
@@ -1,11 +1,137 @@
...
+use status::StatusCode;
+use std::collections::HashMap;
+use std::io::{Error, Read};
+use typemap::Key;
+use url::{form_urlencoded, UrlParser};
+
+type QueryStore = HashMap<String, Vec<String>>;
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct Query(QueryStore);
```
Ryman: If this is going to get reused for both of these, then we should probably rename it. I think we'll have to leave a re-export for `pub type Query = NewTypeName` incase downstream code is depending on this typename.

Perhaps `MultiMap`?

src/urlencoded.rs
```diff
@@ -1,11 +1,137 @@
...
+    ///
+    /// fn main() {
+    ///     let mut server = Nickel::new();
+    ///     server.post("/a", middleware! { |req, res|
+    ///         let post = try_with!(res, req.post());
+    ///         return res.send(format!("Post: {:?}", post))
+    ///     });
+    /// }
+    /// ```
+    fn post(&mut self) -> Result<&Query, (StatusCode, Error)>;
```
Ryman: I'd suggest a rename to `form_body` here which would be more inline with express. cc @cburgdorf

src/urlencoded.rs
```diff
@@ -1,11 +1,137 @@
...
+impl Key for FormBodyParser {
+    type Value = Query;
+}
+
+impl<'mw, 'conn, D> Plugin<Request<'mw, 'conn, D>> for FormBodyParser {
+    type Error = Error;
+
+    fn eval(req: &mut Request<D>) -> Result<Query, Error> {
+        let mut buf = String::new();
+        try!(req.origin.read_to_string(&mut buf));
```
Ryman: This and the json equivalent should likely depend on a `Plugin` which just reads and caches the body (as `Vec<u8>`), else if a user tries to parse json, fails, then parses the form, I think it would fail. (Ultimately this is a failure of the current lack of robust api on `Request`).

It would be good to capture this in an integration test here: https://github.com/nickel-org/nickel.rs/tree/master/tests/examples

Ryman:
I left some comments inline. I was surprised to see we didn't have a test for the form example (I guess I missed that one) as it should have failed when you changed the template. Adding an example-test to [here](https://github.com/nickel-org/nickel.rs/tree/master/tests/examples) is something I'd like to see before merging this.

examples/form_data/form_data.rs
Ryman: Also, it would be good to see what matching on the error looks like, I know you raised the point about returning `(StatusCode, Error)` in the other thread. Perhaps it's better to introduce a new error type? 

How should things be handled if the method is called when no form-data is available, or if it was posted as json/<other-data-format> instead?

Kilte: > I guess there's no easy way to get this working with serialization...

Let's leave it for future )

> Perhaps it's better to introduce a new error type? 

Yes, it would be better to provide more information about an error.

src/lib.rs
Kilte: OK

src/urlencoded.rs
Kilte: > then we should probably rename it.

Why? Query string may be used in URL and request body both. I don't see a reason to do this.
Finally, MultiMap is not clear IMHO. Maybe QueryMap? But is it worth?

src/urlencoded.rs
Kilte: AFAIK, express has no `form_body`, but there is [body-parser](https://github.com/expressjs/body-parser#bodyparserurlencodedoptions) middleware, which replaces req.body. In order to be more inline with express, we need to have a similar middleware. It's better to rename it though.

src/urlencoded.rs
Kilte: > if a user tries to parse json, fails, then parses the form

It should depend on the Content-Type header.
But you're right, it would be nice to have this plugin.

GitCop:

 * Commit: f13a9803eeeb078f085c87d0c5b985d6abb3eee4

---

src/body_parser.rs
```diff
@@ -0,0 +1,124 @@
...
+use std::error::Error as StdError;
+use std::fmt;
+use std::io::{self, ErrorKind, Read};
+use typemap::Key;
+use urlencoded::{self, Query};
+
+struct BodyReader;
+
+impl Key for BodyReader {
+    type Value = String;
```
Ryman: I think this might be more useful in future to be `Vec<u8>` and be public, but I'm happy to leave it like this as we should probably just clean up the `Request` api before making any more externalities based on it.

src/body_parser.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.get_ref::<FormBodyParser>().map_err(|e| (StatusCode::BadRequest, e))
+    }
+}
+
+pub trait JsonBody {
+    fn json_as<T: Decodable>(&mut self) -> Result<T, io::Error>;
+}
+
+impl<'mw, 'conn, D> JsonBody for Request<'mw, 'conn, D> {
+    fn json_as<T: Decodable>(&mut self) -> Result<T, io::Error> {
```
Ryman: Can you add a FIXME here to update the error type on `json_as` when we rename/break it in future? Would be good to capture that possible Parsing error rather than a generic `io::Error`.

src/urlencoded.rs
```diff
@@ -1,20 +1,19 @@
...
 use url::{form_urlencoded, UrlParser};
 
-type QueryStore = HashMap<String, Vec<String>>;
+pub type QueryStore = HashMap<String, Vec<String>>;
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct Query(QueryStore);
 
 impl Query {
+    pub fn new(store: QueryStore) -> Query {
```
Ryman: Could probably just change the return type of `parse` as it seems to be the only usage?

tests/examples/form_data.rs
```diff
@@ -0,0 +1,73 @@
...
+</html>
+";
+
+#[test]
+fn form_data() {
+    run_example("form_data", |port| {
+        // Display form
+        let url = format!("http://localhost:{}/", port);
+        let s = read_url(&url);
+        assert_eq!(s.trim(), FORM.trim());
```
Ryman: I think we can reduce these tests to be a little less precise, so if we update the template in future we don't have to wrestle with fixing whitespace here.

```rust
assert!(s.contains(r#"<form action="confirmation">"#), "response didn't have a form")
```

tests/examples/form_data.rs
```diff
@@ -0,0 +1,73 @@
...
+           .body(body)
+           .send()
+           .unwrap();
+        let s = read_body_to_string(res);
+        assert_eq!(s.trim(), CONFIRMATION.trim());
+
+        // POST failed (invalid content-type)
+        let ref mut res = Client::new().post(&url).send().unwrap();
+        let s = read_body_to_string(res);
+        assert_eq!(s.trim(), "Bad Request");
```
Ryman: Can you assert the StatusCode returned rather than matching the string?

src/urlencoded.rs
Ryman: My concern is for users reading the return type in their rustdoc output, it may be confusing if they see `form_data(..) -> &Query` rather than `form_data(..) -> &FormParams`. Or in the case where the compiler is throwing type errors with `Query` in it when they think they are not dealing with queries.

src/urlencoded.rs
Ryman: Yeah, I just don't think using `body` is a good idea yet until we can handle more things, hence `form_body` :P Sorry for causing confusion, thanks for taking care of it :+1:

tests/examples/form_data.rs
Ryman: Also changing the template from a definition list to a parameterized sentence would help for checking the `POST success` case.

examples/form_data/form_data.rs
```diff
@@ -14,15 +13,16 @@ fn main() {
...
+        data.insert("lastname", form_data.get("lastname").unwrap_or("What's your last name?"));
+        data.insert("phone", form_data.get("phone").unwrap_or("Give me your phone or I will give you mine!"));
+        data.insert("email", form_data.get("email").unwrap_or("I want your email!"));
```
Ryman: The tests don't really exercise the `unwrap_or`s here. 

Compared to the json api, where you could state your required fields by being non-`Option` types, this api looks like it's going to require a lot of extra validation by the user.

Maybe we can add a method `require_keys()` to this, which can return an error with a list of the missing keys, e.g.:
```rust
if let Err(missing) = form_data.require_keys(["firstname", "lastname", "email"]) {
    // rerender the form with missing fields highlighted
}

File: src/body_parser.rs [hunk 12]
```diff
@@ -0,0 +1,124 @@
...
+    fn from(err: io::Error) -> BodyError {
+        BodyError::Io(err)
+    }
+}
+
+impl StdError for BodyError {
+    fn description(&self) -> &str {
+        match *self {
+            BodyError::Io(ref err) => err.description(),
+            BodyError::ContentType => "Invalid content type"
```
Ryman: I think "Invalid" is the wrong word here, as it could be totally valid but just not be the one we expected to parse. Perhaps "Wrong", "Incorrect" or "Unexpected" is better? (I think the variant name should also be updated to `WrongContentType` or `UnexpectedContentType`.

src/body_parser.rs
Ryman: Additional FIXME: do the content-type check.

GitCop:

 * Commit: f13a9803eeeb078f085c87d0c5b985d6abb3eee4

 * Commit: be2ef676d7f59d2367d1a4f5fad3659dfab33e13

---

examples/form_data/form_data.rs
Kilte: > Maybe we can add a method require_keys()

I don't think that is a good idea. It would be really nice to get this working with serialization as you pointed above:
```rust
let foo: SomeType = try_with!(req, form_data.parse())
```

Ryman:
@cburgdorf, @SimonPersson I'm fairly happy with how this currently is. Do you have any suggestions?

@Kilte Would you mind squashing your fixup commits when you get a chance?

tests/examples/form_data.rs
```diff
@@ -0,0 +1,34 @@
...
+                .header(ContentType::form_url_encoded())
+                .body(body)
+                .send()
+                .unwrap();
+            read_body_to_string(res)
+        };
+        let s = f(r#"firstname=John&lastname=Doe&phone=911&email=john@doe.com"#);
+        assert!(s.contains(r#"John Doe 911 john@doe.com"#), "response didn't have an expected data");
+        let s = f(r#""#);
+        assert!(s.contains(r#"First name? Last name? Phone? Email?"#), "response didn't have an expected data");
```
Ryman: If you have time, it would be good to split this test into several to make intention clearer and for easier debugging in the failure case. (these assertions kind of get lost in the flow of things, especially when the test finishes with testing a failure case)

SimonPersson:
@Ryman: Looks good to me as well :+1:

Kilte:
> Would you mind squashing your fixup commits when you get a chance?
Of course, done!

cburgdorf:
yep, let's go for it 👍 Thanks for pushing it further!

Ryman:
@homu r+

Ryman:
@Kilte Thanks for the contribution!

Issue #328, opened 2016-04-11
Testing Guide


pallan:
Really like the framework but am getting stuck on how to test.  I see the examples have tests in the repo, but I can't figure out how to write tests for my on repo.  Please help!

Ryman:
Sorry for the delayed response, the in-repo tests are pretty much a hack and that's been *good enough* for the other time's I'd personally needed to test something using nickel. 

I wanted to put together an example of something a bit saner but got sidetracked a number of times.

I've *finally* opened #334 with at least a sketch of how you can get started with your own basic testing setup. If you could give any feedback on the suitability of that PR for your needs, that would be much appreciated!

PR #334, opened 2016-05-04
feat(server): return `ListeningServer` from `Nickel::listen`

Ryman:
Related to #328.

The tests within this repo are a bit of a hack (require shelling out to subprocesses involving cargo), with this change we can instead make it *easier* to integration test a server (or multiple instances of one) all within the same process which should be easier and faster.

The code within the integration test example would get tested using a normal `cargo test`, but since it's in our example files I had to use a workaround to get it compiled directly into our integration test runner.

BREAKING CHANGE: This changes the return type of `Nickel::listen`, you can add an `unwrap()` to the call to maintain previous semantics.

### DIFF
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -1,9 +1,10 @@
 use std::net::ToSocketAddrs;
 use std::time::Duration;
 use std::env;
+use std::error::Error as StdError;
 use router::{Router, HttpRouter, Matcher};
 use middleware::{MiddlewareStack, Middleware, ErrorHandler};
-use server::Server;
+use server::{Server, ListeningServer};
 use hyper::method::Method;
 use hyper::status::StatusCode;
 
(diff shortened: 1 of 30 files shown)


src/nickel.rs
```diff
@@ -178,18 +179,17 @@ impl<D: Sync + Send + 'static> Nickel<D> {
...
-    ///
     /// # Examples
-    /// ```{rust,no_run}
+    /// ```rust
     /// use nickel::Nickel;
     ///
     /// let server = Nickel::new();
-    /// server.listen("127.0.0.1:6767");
+    /// let listening = server.listen("127.0.0.1:6767").expect("Failed to launch server");
+    /// println!("Listening on: {:?}", listening.socket());
```
SimonPersson: Should this print be here?

SimonPersson:
Exciting! This looks nice to me, but I'm not known to do a lot of testing so let's see if we can get some user feedback.

src/nickel.rs
Ryman: It was to show the use of `socket()` to get the address, but I guess this example will print twice (since the options were not passed to disable printing). So I think it's fine, but I should make the intention more explicit perhaps?

SimonPersson: Okay, I see. As you wish, I think we can just leave it like this.

Ryman:
Rebased on master

euclio:
Is this still in flight?

Ryman:
@euclio just rebased. Do you have any suggested edits / concerns? If not I will try to merge this tomorrow.

euclio:
Nope, looks good to me. Thanks for your work!

euclio:
@Ryman friendly ping :)

Ryman:
@homu r+

PR #335, opened 2016-05-06
feat(options): add an option for thread count

tanriol:
Add an option for server thread count from #309 to the new Options API.

### DIFF
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -19,11 +19,15 @@ use default_error_handler::DefaultErrorHandler;
 /// use nickel::{Nickel, Options};
 /// let mut server = Nickel::new();
 ///
-/// // Don't print to stdout when starting the server.
-/// server.options = Options::default().output_on_listen(false);
+/// // Don't print to stdout when starting the server
+/// // and force using 8 threads.
+/// server.options = Options::default()
+///                      .output_on_listen(false)
+///                      .thread_count(Some(8));
 /// ```
 pub struct Options {
     output_on_listen: bool,
+    thread_count: Option<usize>,
 }
 
 impl Options {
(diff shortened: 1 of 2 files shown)


tanriol:
Travis failed on Rust 1.7 with SSL because openssl-0.7.11 allowed bitflags-0.6 which requires Rust 1.8 (augmented assignment).

Ryman:
The version bump is quite a shame :(

In the previous thread, we discussed leaving the default to hyper, but there's been a number of people asking for this feature who are suffering thread exhaustion, so perhaps we should raise the default limit here.

src/nickel.rs
```diff
@@ -19,11 +19,12 @@ use default_error_handler::DefaultErrorHandler;
 /// use nickel::{Nickel, Options};
 /// let mut server = Nickel::new();
 ///
-/// // Don't print to stdout when starting the server.
-/// server.options = Options::default().output_on_listen(false);
+/// // Don't print to stdout when starting the server and force using 8 threads.
+/// server.options = Options::default().output_on_listen(false).thread_count(Some(8));
 /// ```
```
Ryman: This line is rather long, can you wrap it?

src/nickel.rs
```diff
@@ -34,12 +35,21 @@ impl Options {
         self.output_on_listen = output;
         self
     }
+
+    /// The number of threads the server should use or `None` for default.
+    ///
+    /// Defaults to `None`.
```
Ryman: We should probably specify what default means (related to my comment in thread, do we still keep to hypers default or do we maintain our own default)

src/nickel.rs
tanriol: Sure, should I add new commits or rebase?

Ryman: I think a rebase would be best in this case as it's a minor change to something that hasn't yet merged. Thanks!

tanriol:
Wrapped lines according to rustfmt hints and added a little detail on the default.

tanriol:
On Travis the same problem; on AppVeyor github timed out in the i686 build.

Ryman:
I've opened https://github.com/nickel-org/nickel.rs/pull/336 to fix the travis failure (another option would be to bump the minimum version, but I'd prefer if we could continue to build on 1.7 in the 0.8 release)

Ryman:
@homu r+

Ryman:
Thanks @tanriol!

PR #336, opened 2016-05-09
fix(deps): force bitflags v0.5 if ssl is enabled

Ryman:
This should allow us to continue building on rustc 1.7 rather than requiring 1.8, 
which is currently the very latest version released. 

We could remove this when we next version bump if we want.

r? @SimonPersson @cburgdorf

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["nickel", "server", "web", "express"]
 
 [features]
 unstable = ["hyper/nightly", "compiletest_rs"]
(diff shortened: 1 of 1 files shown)


tanriol:
Does not work because Cargo allows both 0.5 and 0.6 to be pulled into the dependency graph.

Ryman:
Ah that's a pain, looks like we'll just have to bump our version requirement..

sfackler:
You can use `cargo update` to control the bitflags version selected:

```
$ cargo update -p bitflags --precise 0.5.0
```

PR #337, opened 2016-05-09
fix(travis): Don't fail the build when minimum rustc version fails

Ryman:
This still keeps the minimum rustc version tested on travis as a canary for when things do inevitably fail in future.


Ryman:
@homu r+

PR #339, opened 2016-05-16
chore(mimes): improve macro definitions

dotdash:
By moving common code out of the match arms, compile times for
optimized builds are reduced by about 25% on my box, and the resulting
binary is almost 500kb smaller.

### DIFF
--- a/src/mimes.rs
+++ b/src/mimes.rs
@@ -21,24 +21,24 @@ macro_rules! mimes {
                 match mt {
                     $(
                         $(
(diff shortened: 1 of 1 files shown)


Ryman:
@homu r+

Thanks @dotdash (for this and the related compiler improvements)! 

Pretty unexpected that something so small can have such a dramatic effect on compilation. I think we should probably pull this out into a separate crate to depend on or move to depend on mimeguess or something.

PR #341, opened 2016-05-28
chore(*): Upgrade hyper and url versions

sfackler:
Also move listen_https to be unconditionally included since openssl
isn't required by hyper anymore.

### DIFF
--- a/src/nickel.rs
+++ b/src/nickel.rs
@@ -5,6 +5,7 @@ use router::{Router, HttpRouter, Matcher};
 use middleware::{MiddlewareStack, Middleware, ErrorHandler};
 use server::Server;
 use hyper::method::Method;
+use hyper::net::SslServer;
 use hyper::status::StatusCode;
 
 //pre defined middleware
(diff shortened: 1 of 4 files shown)


src/nickel.rs
```diff
@@ -273,45 +247,64 @@ mod ssl {
...
     /// use hyper::net::Openssl;
     ///
+    /// # #[cfg(feature = "ssl")]
     /// # fn main() {
     /// let server = Nickel::new();
     /// let ssl = Openssl::with_cert_and_key("foo.crt", "key.pem").unwrap();
     /// server.listen_https("127.0.0.1:6767", ssl);
     /// # }
+    /// # #[cfg(not(feature = "ssl"))]
+    /// # fn main() {}
```
Ryman: Is there an example we can give that works without hyper's `ssl` feature? It seems like our own ssl feature is only for this doctest and the ssl example now. That seems like it could cause confusion, would renaming our feature to `openssl` be accurate in your opinion? 

I guess this is related to your work on other ssl implementations that aren't openssl? Should we be recommending those to users now? (Thanks for the work on those crates btw!) 

(This also fails on Travis due to unused imports: `nickel::Nickel`)

src/urlencoded.rs
```diff
@@ -30,20 +30,20 @@ impl Params {
...
-        AbsoluteUri(ref url) => f(url.query.as_ref()),
-        AbsolutePath(ref s) => UrlParser::new().parse_path(&*s)
-                                                // FIXME: If this fails to parse,
-                                                // then it really shouldn't have
-                                                // reached here.
-                                               .ok()
-                                               .and_then(|(_, query, _)| f(query.as_ref())),
+        AbsoluteUri(ref url) => f(url.query()),
+        AbsolutePath(ref s) => {
+            let base = "f://".parse().unwrap();
```
Ryman: This seems a bit random, can you add a comment (or link) to why it's got some non-http protocol?

src/nickel.rs
sfackler: Originally TLS support in Hyper was hardcoded to rust-openssl, but that was changed so that Hyper could deal with any TLS backend through a couple of traits. For historical reasons, Hyper's `ssl` feature is the one that enables the implementations of those traits for rust-openssl, but it should really be renamed at some point.

Because of this, I think the the nickel feature can be removed entirely now, actually - Nickel doesn't need to care about what `SslServer` implementations exist, other than that *some* implementation is being given to `Nickel::listen_https`.

That does complicate this example a bit, though - it should maybe be just tagged `ignore` at that point so it doesn't need to compile.

As far as what ssl implementations people should use, rust-openssl's probably the right thing to recommend for unixy platforms other than OSX. OSX ships with a copy of OpenSSL, but it's *super* old and Apple is fairly aggressively trying to kill it off. In particular, El Capitan no longer includes OpenSSL headers, which prevents rust-openssl from compiling so you have to pull a copy in with something like Homebrew.

The bindings for OSX's native Secure Transport APIs are in a pretty decent place I think, I'd probably recommend that on OSX for most simple use cases.

On Windows, OpenSSL is even more of a nightmare to get set up, but the bindings to SChannel, Windows's native SSL API, still need a lot of work, so you're pretty much stuck with OpenSSL there, unfortunately.

src/urlencoded.rs
sfackler: The protocol here doesn't actually matter, we just need some arbitrary base URL to parse the absolute path off of since rust-url dropped support for parsing subcomponents of the URL (is that right, @SimonSapin?).

SimonSapin: rust-url 1.0 is almost a rewrite. I removed a lot of code after changing the data structure in order to get anything to compile, and added things back / rewrote them gradually. I neglected to add standalone path parsing back in 1.0 or 1.1.

We could add it back in 1.2, but it’s not clear to me that that you want rust-url’s exact parser for this. The WHATWG URL spec is primarily designed for clients (typically, parsing a `<a href="...">` HTML attribute). Does an HTTP server want to every edge case the same way? For example, percent-encoding (not decoding) some punctuation/delimiters, normalizing `\` to `/`, normalizing `.` and `..` path segments?

Ryman: In this specific case, we can probably just split the string `s` on `?`.

 

> Does an HTTP server want to every edge case the same way? For example, percent-encoding (not decoding) some punctuation/delimiters, normalizing \ to /, normalizing . and .. path segments?

For this, I'm not so sure, perhaps we should raise the issue in `hyper` as it's already using `Url` for a variant in [`RequestUri`](http://hyper.rs/hyper/v0.9.6/hyper/uri/enum.RequestUri.html). cc @seanmonstar

Ryman:
Rebased and merged as https://github.com/nickel-org/nickel.rs/commit/b77e4ca9a5544a742ec2fb894b85b6b736239656. Thanks @sfackler !

PR #349, opened 2016-06-21
feat(hyper): add re-export for Method

bozaro:
I use nickel-0.8.1 for server side and h hyper-0.9.3 for client side.
Unfortunatelly I can't add route for HEAD method without this change :(

### DIFF
--- a/src/favicon_handler.rs
+++ b/src/favicon_handler.rs
@@ -3,12 +3,12 @@ use std::path::{PathBuf, Path};
 use std::io::Read;
 
 use hyper::uri::RequestUri::AbsolutePath;
-use hyper::method::Method::{Get, Head, Options};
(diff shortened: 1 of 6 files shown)


Ryman:
I'm 👍 for this, I'd love to actually remove hyper from most of the public api so we don't need to be so tied to its version number bumps. Perhaps we should consider grouping some of them under `hyper` though and only reexport some of them at the top levels e.g. `nickel::hyper::less_important_import`? (I'm not sure on this though, but I think I've seen some other libraries do something similar?)

bozaro:
I do not have a strong opinion on the matter. I use nickel only about three days :)

Ryman:
Closing this as it looks like `hyper` is now exported so it should be accessible through `nickel::hyper::Method`

PR #356, opened 2016-07-09
chore(*): Upgrade hyper and url versions

Ryman:
Also move listen_https to be unconditionally included since openssl
isn't required by hyper anymore.

This is a rebase of #341 with some minor changes.


Ryman:
@homu r+

PR #358, opened 2016-07-11
fix(error_handling): introduce IntoError trait to reduce coherence issues

Ryman:
Rebase-ish of https://github.com/nickel-org/nickel.rs/pull/304

I don't think this will actually be that breaking, but since it is breaking I'd like to get it in with the next release.

The related PR came up during https://github.com/nickel-org/nickel-postgres/pull/17#issuecomment-221746314 as well which was getting coherence problems which this should hopefully fix.

cc @flosse @hjr3


hjr3:
This is great! The examples are clear and I do not see any problems.

PR #364, opened 2016-07-23
 feat(serde): add support for Serde serialization

### LINKED ISSUE #345: Support Serde
It'd be great to be able to support `serde` instead of `rustc_serialize` for compatibility with the many crates that use `serde` as the serialisation framework of choice.

If this isn't on anyone else's radar and isn't too difficult I might pick it up. I'm keen to dig in to `nickel` anyways 😄

KodrAus:
Closes #345 

This is a first-pass attempt at adding support for `serde` serialisation.

The feature is called `with-serde`, so to use call:

```
rustup run stable cargo build --features with-serde --no-default-features
```

`rustc_serialize` is the default choice, so this shouldn't break existing implementations by default.

`rustc_serialize` has served us well, but it's clear that `serde` is the framework to support in the longterm. I wouldn't expect this feature to land for a little while yet; `serde` is working through a new breaking release which would make sense to support (most of that ugly manual serialisation in the examples will go away), but will also need to be updated in the [PR](https://github.com/nickel-org/rust-mustache/pull/40) for `mustache`.

A few things on my mind:
- Serialising to `serde_json::Value`, and then from `Value` to a `String` when building a response seems wacky. I'm thinking of adding a struct `Json<'a T: Serialize>(&'a T)` (could get away without the borrow if need be) that can implement `Responder`, to set mime properly, but avoid that extra hop.
- Do the cfg attributes better now that I have a bit of a better idea how `nickel` works.
- Make the cruft I've introduced to the example tests a bit nicer (it's not really representative of how people would use it anymore, because nobody is going to make their app support both serialisation frameworks). Some feedback here would be great!

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,4 +13,4 @@ install:
   - cargo -vV
 build: false
 test_script:
-  - cargo test --verbose --no-default-features
+  - cargo test --verbose
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,9 @@ readme = "README.md"
 keywords = ["nickel", "server", "web", "express"]
 
 [features]
+default = ["rustc-serialize"]
 unstable = ["hyper/nightly", "compiletest_rs"]
+with-serde = ["serde", "serde_json", "mustache/serde"]
 ssl = ["hyper/ssl"]
 
 [dependencies]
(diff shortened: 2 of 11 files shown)


KodrAus:
Hopefully I can fix that crummy history, must've come in when I rebased this morning...

jolhoeft:
I recreated this in PR #435

PR #371, opened 2016-08-07
fix(parameters): Issue regarding parameters that contain a "."

### LINKED ISSUE #360: dots in param break routing?
Hi folks,

I have a routing to `/node/:hostname`. Now if I go to `/node/foobar` it routes fine, if it includes one dot like `/node/foo.bar/` it also works but as soon as there is a second dot there like `/node/foo..bar` or `/node/foo.bar.` I get routed to the 404 Not Found.

Is this expected? I'm asking because actually I need to get an ip address as a param which has 3 dots :)

BenKnigge:
resolves #360

### DIFF
--- a/src/router/into_matcher.rs
+++ b/src/router/into_matcher.rs
@@ -21,10 +21,10 @@ lazy_static! {
 pub static FORMAT_PARAM:      &'static str = "format";
 // FIXME: Once const fn lands this could be defined in terms of the above
 static FORMAT_VAR:            &'static str = ":format";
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 4a734bc9e0e1454a5190bd6a9980e7c2d2fbcb25

 * Commit: 2f52b0f6d1ce82e9b29f1caea812784ce8696762

 * Commit: 0274a0e861f5f8ed76d797ed21c3562e0cfcc5ea

---

PR #373, opened 2016-08-07
fix(router): issue regarding matching "."

BenKnigge:
resolves issue #360


BenKnigge:
Based on the test failure it appears that the issue regarding "." in parameters is deliberate, as "." is being used as a parameter delimiter which causes the test examples::routing::parameterised_path_with_format. Has this issue been addressed previously ?

SimonPersson:
Yeah, @Ryman, was there any discussion regarding that? It seems a bit unexpected to me that we automatically extract a format. It would make more sense to me to have that as a special case, requiring the user to write the route as a regex, or put in a middleware that extracts the format.

Issue #399, opened 2017-05-03
Compile causes warnings about macros


kpcyrd:
Compiling my project results in a bunch of warnings:

```
warning: code relies on type inference rules which are likely to change
  --> src/main.rs:83:21
   |
83 |       server.get("/", middleware! { |_, res|
   |  _____________________^ starting here...
84 | |         let mut data = HashMap::new();
85 | |         data.insert("name", "user");
86 | |         return res.render("templates/index.html", &data);
87 | |     });
   | |_____^ ...ending here
   |
   = note: #[warn(resolve_trait_on_defaulted_unit)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #39216 <https://github.com/rust-lang/rust/issues/39216>
   = note: this error originates in a macro outside of the current crate

warning: unreachable expression
  --> src/main.rs:83:39
   |
83 |     server.get("/", middleware! { |_, res|
   |                                       ^^^
   |
   = note: #[warn(unreachable_code)] on by default

warning: code relies on type inference rules which are likely to change
  --> src/main.rs:89:28
   |
89 |       server.get("/e/:hash", middleware! { |req, res|
   |  ____________________________^ starting here...
90 | |         let hash = req.param("hash").unwrap();
91 | |
92 | |         let mut data = HashMap::new();
93 | |         data.insert("hash", hash);
94 | |         return res.render("templates/view.html", &data);
95 | |     });
   | |_____^ ...ending here
   |
   = note: #[warn(resolve_trait_on_defaulted_unit)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #39216 <https://github.com/rust-lang/rust/issues/39216>
   = note: this error originates in a macro outside of the current crate

warning: unreachable expression
  --> src/main.rs:89:48
   |
89 |     server.get("/e/:hash", middleware! { |req, res|
   |                                                ^^^
   |
   = note: #[warn(unreachable_code)] on by default
```

This happens with `router!` as well. My code is very similar to the examples on http://nickel.rs, I'm not sure if I'm doing something wrong.

davidpodhola:
Same here, related to https://github.com/rust-lang/rust/issues/39216

jolhoeft:
This appears to be causing four tests to fail during cargo test.

jolhoeft:
[...]

Issue #389, opened 2016-12-07
Early return inside middleware! macro


mmacedoeu:
I need to return early if preconditions are not met, but tuple response is not accepted if I use return:

```rust
fn main() {
  let mut server = Nickel::new();
  server.post("/claim",
    middleware!{ |request, response|
      let condition : bool = try_with!(response, {handle_preconditions().map_err(|e| (StatusCode::BadRequest, e))});

     if !condition {
       return (StatusCode::PreconditionRequired, "invalid state".to_string()); // this is not compilling
       // expected type `std::result::Result<nickel::Action<nickel::Response<'mw, _>, //nickel::Response<'mw, _, hyper::net::Streaming>>, nickel::NickelError<'mw, _>>`
//    = note:    found type `(hyper::status::StatusCode, std::string::String)`
     }

      let good : bool = try_with!(response, {handle_processing().map_err(|e| (StatusCode::BadRequest, e))});

      if good {
         (StatusCode::Accepted, "Fine".to_string())
      } else {
         (StatusCode::NotAcceptable, "Not Fine".to_string())
      }

    });
    server.listen("127.0.0.1:6767").unwrap();
}
```

It is expecting a return of type `std::result::Result<nickel::Action<nickel::Response<'mw, _>, nickel::Response<'mw, _, hyper::net::Streaming>>, nickel::NickelError<'mw, _>>`

How can I early return ?

mattlknight:
I'm not very good at offering advice, but im currently banging my head against Nickel.rs and it's macros.

I believe this might help you.

In your file, use 
`use nickel::MiddlewareResult;`
and then for your early return use `return response.send(put your stuff here);`
Ex:  
`return response.send((StatusCode::Accepted, "Fine".to_string()));`

Kilte:
I guess this should work:
```rust
return (response, (StatusCode::PreconditionRequired, String::from("invalid state")));
```

jolhoeft:
I figured this out while sorting out #399. Basically the middleware macro is somewhat fragile. It looks like it takes a closure, but is actually taking a block. The block must return something implementing Responder. Note that since it is a block, the return statement interacts with the macro expansion in unexpected ways. To make the original example work, you need to not use return, and instead nest the if statements:

```rust
fn main() {
    let mut server = Nickel::new();
    server.post("/claim", middleware!{ |request, response|
[...]

PR #414, opened 2017-11-19
fix(examples): Fix examples triggering issue #399

jolhoeft:
Fix the examples that were returning MiddlewareResults in the
middlware! macro. A couple could be changed to return a Responder, but
most needed to be reimplemented as seperate functions.

Updated the middleware! macro documentaion to note the limitations
from issues #399 and #389.


jolhoeft:
Unless there are loud objections, I am going to merge this quickly, because it repairs our continuous integration tests, which have been failing for months.

jolhoeft:
Ok, trying again, with the SSL testing removed, (broken months ago see #415).

PR #428, opened 2018-02-22
fix contributing.md formatting

DevOrc:
This adds a space before headers in [contributing.md](https://github.com/nickel-org/nickel.rs/blob/master/contributing.md). Currently the headers don't have a space so they look like this:
##HeaderExample
The header should have a space in between the "##" and the text like this:
## Header Example


jolhoeft:
Thank you

PR #435, opened 2018-11-29
feat(serde): replace rustc_serialize with serde

jolhoeft:
Addresses #345


jolhoeft:
My testing has worked, and there are no comments, so I am merging and pushing to crates.io.

Issue #405, opened 2017-07-12
StaticFilesHandler doesn't handle encoded URL for whitespace


tungel:
If a request to server is `test%20hi%20foo%20bar/python.jpg` , it seems StaticFilesHandler will just literally take it as the path to the file system (which doesn't exist). It should decode the request URL into `test hi foo bar/python.jpg`  instead.

PR #457, opened 2020-07-25
fix(static_file_handler): Support for blank spaces and percent encodi…

hdf1986:
# Summary
Adds handling for whitespaces in url for static files handler. 

# Issue
#405 

# Note
Im really new at Rust, and even more with ownership and low level memory handling, so probably i've made some mistakes in the way i changed this to a String. I'm more than open to hear suggestions on how to do this any better.
Fix inspired by https://github.com/lawliet89/nickel.rs/commit/f1c03e27c47366c974322e9db2904edd0ffcb9de

### DIFF
--- a/src/static_files_handler.rs
+++ b/src/static_files_handler.rs
@@ -45,13 +45,15 @@ impl StaticFilesHandler {
         }
     }
 
-    fn extract_path<'a, D>(&self, req: &'a mut Request<D>) -> Option<&'a str> {
+    fn extract_path<'a, D>(&self, req: &'a mut Request<D>) -> Option<String> {
         req.path_without_query().map(|path| {
-            debug!("{:?} {:?}{:?}", req.origin.method, self.root_path.display(), path);
+            let percent_decoded_path = percent_encoding::percent_decode(path.as_bytes()).decode_utf8().unwrap();
+
+            debug!("{:?} {:?}{:?}", req.origin.method, self.root_path.display(), percent_decoded_path);
 
             match path {
-                "/" => "index.html",
-                path => &path[1..],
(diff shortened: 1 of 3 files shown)


jolhoeft:
This looks reasonable to me. We are now allocating a new String with each call, but I'm willing to pay that price. Could you add a unit test to verify percent decoding is working correctly?

hdf1986:
Thanks for the feedback @jolhoeft. I extracted the code to a fn to be able to test it easier, and included a few sample paths to the test

src/static_files_handler.rs
```diff
@@ -45,13 +47,15 @@ impl StaticFilesHandler {
         }
     }
 
-    fn extract_path<'a, D>(&self, req: &'a mut Request<D>) -> Option<&'a str> {
+    fn extract_path<'a, D>(&self, req: &'a mut Request<D>) -> Option<String> {
         req.path_without_query().map(|path| {
-            debug!("{:?} {:?}{:?}", req.origin.method, self.root_path.display(), path);
+            let percent_decoded_path = percent_decode_path(path).unwrap();
```
jolhoeft, line 52: Sorry about losing track of this. I am a little concerned about the `unwrap` here. If that gets triggered, the current thread will be killed. That may be an opening for a DOS, where a series of ill formed URLs exhausts all the threads. We should probably change how we are handling threads to avoid that, but I want to defer that until after I get the async migration completed.

This should probably be changed to return `None` if the path cannot be decoded.

hdf1986: Got it, will do. I might take some time to do the change since im busy with some personal stuff, but i'll make sure to do the change as soon as i have some free time. Thanks!

PR #462, opened 2022-04-23
chore(*): update dependencies and Rust edition

damfle:
I was unable to create an example on the latest version of Rust.
Now everything works fine (tests and examples) and is up to date.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
 
 name = "nickel"
-version = "0.12.0"
+version = "0.12.1"
 authors = [ "Christoph Burgdorf <christoph@thoughtram.io>",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [package]
 
 name = "nickel"
-version = "0.12.0"
+version = "0.13.0"
```
jolhoeft, line 4: Thank you for this. Let's make the version "0.12.1". I've not yet released 0.12 to crates.io yet, so there is no backward compatibility to break yet.

damfle: Done. Amended as this is only a constant change

jolhoeft:
Thank you!

